PR #1, opened 2016-11-29
Chrom convert

jfear:
For dm6, replaces the appending of "chr" with the chrom_convert script. Also moves the wrappers submodule branch to current master.

Re-ran reference.snakefile and everything builds cleanly.


daler:
Awesome.

PR #3, opened 2016-12-05
Config restructure

daler:
There were multiple moving parts here I wanted to fix, so this is kind of a monolithic PR.

First, some directory structure cleanup:

- There wasn't a good reason to have the reference.snakefile in the references directory. There were some useful bits I wanted to use over in rnaseq.snakefile (which are now living in common.py), and the only good way to make that work was either to do sys.path manipulation or to commit everything to lcdblib. I'd rather not have lcdblib have code that's tightly coupled to the snakefiles.

- The various post-processing modules (dm6.py, hg19.py, etc) have been moved to a postprocessing directory from which references.snakefile can import as needed.

- the references dir can be removed once the readme is migrated to proper docs

Then, some conda stuff:

- if you use the development version of snakemake (with my conda env creation fixes), it it now possible to run the references workflow with nothing but snakemake, lcdblib, and conda installed. As long as you use the `--use-conda` arg. This makes wrappers run in their own environments that are automatically created on-the-fly. There's also a new `envs` directory that will store environments for workflows. Right now it's just `references_env.yml` which as dependencies that aren't yet a wrapper (kallisto index) or aren't worth a wrapper (gtfToGenePred). Rules have `conda: "envs/references_env.yml"` when they need this.

- I pushed a change to the wrappers where bowtie2 and hisat2, which use the indexes_from_prefix and prefix_from_index from lcdblib.aligners, now install lcdblib from git via pip in their environments.

And finally some config stuff:

- The config file structure was difficult to work with. While the flat structure was convenient for incrementally adding new assemblies and tags, it was annoying to pull out needed information. So now it's organized to be more hierarchical. I think this makes it a little easier to follow the directory structure, but importantly it allows you to select a tag and type for things like rnaseq alignment (rather than needing to grok the directory structure and then write the path you need). I think the new layout helps avoid errors, too: I realized there was an extra dicty transcriptome block while reformatting it.

- common.py has the references_dict function to convert config file into something convenient to use within workflows. The doctest for that function illustrates how it works.

- I think the references directory structure could use a little more cleanup so that the config file more closely resembles the directory structure. But this PR does most of the hard work towards getting there.


common.py
```diff
@@ -0,0 +1,241 @@
...
+
+    for url, tmpfile in zip(urls, tmpfiles):
+        shell("wget {url} -O- > {tmpfile} 2> {outfile}.log")
+
+    func(tmpfiles, outfile, *args)
+
+    for i in tmpfiles:
+        if os.path.exists(i):
+            shell('rm {i}')
+
```
jfear: I don't know if it is worth worrying about, but it may be a good idea to wrap the download and post-processing in a try/except and then put the cleanup in a finally.

common.py
```diff
@@ -0,0 +1,241 @@
...
+    ...          'chromsizes': '/data/dm6/fasta/dm6_r6-11_transcriptome.chromsizes',
+    ...          'kallisto': '/data/dm6/kallisto/dm6_r6-11_transcriptome.idx',
+    ...          },
+    ...     },
+    ... }), d
+    >>> os.unlink('tmp')
+
+    With this new dictionary, other parts of the config or snakemake rules can
+    access the files by d[assembly][tag][type].
+    """
```
jfear: much nicer!!

common.py
daler: That's a good idea.

daler:
For the concatenated genome, you can just put the urls in a list. The hard/annoying part would be writing the postprocessing function.

jfear:
I am going to go through the code again in the AM. I really need to fully understand common.py.

common.py
```diff
@@ -0,0 +1,243 @@
...
+    else:
+        if isinstance(post_process, dict):
+            name = post_process.get('function', post_process)
+            args = post_process.get('args', ())
+            if isinstance(args, str):
+                args = (args,)
+        elif isinstance(post_process, str):
+            name = post_process
+            args = ()
+
```
jfear: Probably useful to add kwargs as well.

common.py
```diff
@@ -0,0 +1,243 @@
...
+        d[assembly] = {}
+        for tag in config['references'][assembly].keys():
+            e = {}
+            for type_, block in config['references'][assembly][tag].items():
+                e[type_] = (
+                    '{references_dir}/'
+                    '{assembly}/'
+                    '{type_}/'
+                    '{assembly}_{tag}.{type_}'.format(**locals())
+                )
```
jfear: Wow, this is super creative use of locals(). I don't think I would have ever come up with that.

common.py
```diff
@@ -0,0 +1,243 @@
...
+                if type_ == 'gtf':
+                    conversions = block.get('conversions', [])
+                    for conversion in conversions:
+                        ext = conversion_extensions[conversion]
+                        e[conversion] = (
+                            '{references_dir}/'
+                            '{assembly}/'
+                            '{type_}/'
+                            '{assembly}_{tag}{ext}'.format(**locals())
+                        )
```
jfear: I am trying to think how to pass parameters for conversions. I can think of a lot of different "conversions" that would be useful to add, but some of them would need genome specific parameters (i.e. gffutils). 

Off the top of my head:

* refflat
* intergenic
* Intergneic Bed
* gtf -> gffutils DB
* gffutils DB -> Bed12
* gffutils -> known splice junctions (bed)
* gene/transcript lengths
* Accession to symbol (i.e. FBgn to symbol)

jfear: What do you think about allowing/using a dict instead of list where key is the conversion id and value parameters to pass?

jfear: Nope, this probably won't work sense you are passing the entire references dict as targets.

daler: True, but with the new config format, a conversion's rule can look up the args/kwargs based on the filename:


```yaml
dm6:
  default:
    gtf:
      url: ""
      conversions:
        gffutilsdb:
          kwargs:
            disable_infer_introns: True  
            disable_infer_genes: True
```

```python
rule gffutilsdb:
    output: '{refdir}/{assembly}/gtf/{tag}.gtf.db'
    run:
        kwargs = config[wildcards.assembly]['gtf'][wildcards.tag]['conversions']['gffutilsdb']['kwargs']
```

jfear:
Are there still changes you want to make before merging?

daler:
Yep -- right now directory structure goes `assembly/type/assembly_tag.extension`, but to match the yaml, it'd be nicer to have `assembly/tag/type/assembly_tag.extension`, or possibly  `assembly/tag/type/assembly.extension` (i.e. no tag in basename).

jfear:
Do you have a preference of having tag in the basename? I am leaning towards keeping it. Playing now.

daler:
No preference -- whatever you think is good.

daler:
I just made some fixes so that the rnaseq.snakefile actually runs. Are you playing with the dir structure? I can tackle that now.

jfear:
No I have not started. I am still trying to get the conda stuff to work 
in the real world.

jfear:
Looks good.

PR #6, opened 2017-01-01
Travis tests

daler:
Now that we have a decent framework for the wrapper tests, it was straightforward to add the tests here.

There's a `test_config.yaml` that gets the small fastqs, fasta, and gtf from lcdb test data.

This is currently failing because of HTTP 500 errors from the anaconda servers (which seems to happen periodically). The solution is to keep restarting the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: python
+install: bash travis-setup.sh
+before_script:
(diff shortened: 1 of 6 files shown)


daler:
anaconda servers are back, this is working now.

PR #10, opened 2017-01-13
Mod fix

jfear:
Added a fix for the mod(mdg4) problem I was having. The question is should we add the fix this far upstream or if it would be better to keep the trans splice events and do the filtering some where else.


jfear:
I added this to the battle testing branch.

PR #11, opened 2017-02-08
Battle test

jfear:
I have changed a few things and have the tests working. Going to poke around and a little and see if there is anything else we want to add. I have been running everything locally, will generate references on biowulf too.

`protected` works, but only changes permissions for files, folders are still writable which could be problematic.

May want to add a rule that does:
* Takes all files
* Calls dirname
* runs chmod 555 on the folders

Or we could just make a rule that runs:

`find /data/LCDB/lcdb-references -type d -exec chmod 555 {} \;`
Then touch some done file.

### DIFF
--- a/postprocess/phix.py
+++ b/postprocess/phix.py
@@ -1,0 +1,9 @@
+from snakemake.shell import shell
+
+def fasta_postprocess(origfn, newfn):
+    shell(
+          "tar -xf {origfn} "
+          "&& cat PhiX/Illumina/RTA/Sequence/WholeGenomeFasta/genome.fa | "
+          "fold -w 80 | "
+          "gzip -c > {newfn} "
+          "&& rm {origfn}")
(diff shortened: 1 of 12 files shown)


postprocess/phix.py
```diff
@@ -0,0 +1,9 @@
+from snakemake.shell import shell
+
+def fasta_postprocess(origfn, newfn):
+    shell(
+          "tar -xf {origfn} "
+          "&& cat PhiX/Illumina/RTA/Sequence/WholeGenomeFasta/genome.fa | "
+          "fold -w 80 | "
```
daler: never knew about `fold`, nice!

daler:
phew!

postprocess/phix.py
jfear: I am sure I have forgotten more bash tools then I remember

PR #14, opened 2017-02-16
bootstrap script

daler:
This provides a way of cloning the repo into a new analysis directory.

The idea is that there's a whitelist (`manifest.txt`) of files in the repo that should be copied, anything not on the list is skipped.

```bash
git clone . tmpdir
cd tmpdir
git remote add upstream git@github.com:lcdb/lcdb-wf.git
git checkout -b newbranch
git rm $files-to-remove  # see below
git commit -m 'deleted stuff'
rsync tmpdir/ destdir/
```
Which files to delete from the temporary cloned repo: anything NOT matched by globs in `manifest.txt`, and  anything  that already exists in the destination dir.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 import os
 import sys
 import argparse
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,0 +1,107 @@
+import os
+import sys
+import argparse
+import tempfile
+from snakemake.shell import shell
+import glob
+
+usage = """
+Clones the repo into a new directory ready for a new analysis.
+
+Specifically:
+
+    - clone current repo to a temp dir
+    - check out a new branch in the tmp repo
+    - git rm everything that isn't listed on manifest.txt and commit on the new branch
+    - add the github repo as "upstream" remote
+    - copy the directory over, but warn about conflicts
+"""
+BLUE = '\033[94m'
+GREEN = '\033[92m'
+YELLOW = '\033[93m'
+RED = '\033[91m'
+END = '\033[0m'
+
+ap = argparse.ArgumentParser()
+ap.add_argument('dest', help='Target directory. Will be created if needed')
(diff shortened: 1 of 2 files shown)


jfear:
Awesome, I was actually thinking about how to do this the other day!!

daler:
I'd like to eventually do a

```bash
curl https://raw.githubusercontent.com/lcdb/lcdb-wf/master/bootstrap.py | python
```

jfear:
Sorry, just trying to think things through. Can you describe your standard workflow/file tree to give me a better picture?

Each of my projects is a git repository. The question I am struggling with is how this fits in. I came up with the following:

1) bootstrap to a subfolder of my project repository
2) bootstrap to start a project folder

I am wondering how much communication you usually do with `upstream`. 

For (1), submodules would not really work because you end up removing `origin` and pointing to `upstream`. Here I think a subtree could work, but this makes pushing `upstream` challenging. 

For (2) you can just add a new `origin`, but would get complicated if you need to make new branches.  

Another option would be to use the github importer, but this would also break `upstream`. 

Perhaps I am missing something simple.

daler:
hah, I was just writing these points in the readme! Though it's a bit complicated to explain.

When bootstrapping to an analysis directory *it's expected to not be a git repo yet*. In fact the script exits if it detects a `.git` dir in the destination.

I don't do much communicating with upstream (github), but a lot with origin (that is, wherever the cloned-from-github repo is living in my home directory). If I make a change in the analysis directory I want to be propagated back to origin, I actually use `meld analysis_dir origin_dir` to do the line-by-line changes to origin by hand. Then back in `origin` I make the commits and push to *its* origin, which is github.

Now all analysis directories that were cloned from the same origin can get the new changes. In practice I use `git merge master --no-ff --no-commit` in the analysis dir and then pick-and-choose which commits I want with `git cola`.

There's probably an easier or more straightforward way of doing this. The overall goal is that fixes and improvements in one analysis dir can be propagated out to all other analysis dirs.

PR #19, opened 2017-03-27
reorganization of tests

daler:
WIP . . . I'll update this when things stabilize.

**Update:**

This is an omnibus PR with substantial changes to the test framework. It includes the following changes:

- adds parallel testing support via `pytest-xdist`. Run pytest with `-n 6` to run in parallel with 6 cores.
- uses new data from `lcdb-test-data`'s `add-chipseq` branch for all wrapper tests
- pytest fixture reorganization. See `wrappers/README.md` for details. Basically fixtures are pulled out of `conftest.py` and put in the respective `test_<wrapper name>.py` file.
- most wrapper tests now use the "tiny" fastq and bam files for faster testing, relying on the full rnaseq.snakefile run as an integration test
- uses buildkite for testing in parallel on local infrastructure. This is currently in a 14-day trial period, so we may have to disable later.
- wrapper tests now default to NOT using `--use-conda`, and a "global" env is used for most tests. For cases like rseqc that use an incompatible environment, we can still use `use_conda=True` in the `run()` call.

I was running into a lot of issues with buildkite with parallel tests. There was something funky with conda not finding symlinks, and I played around a lot with installing miniconda or building fresh environments or not. Finally, using a "global" environment combined with 6-way parallelized tests gets the wrapper tests down from ~20 mins to ~4 mins, and with a final setup + wrapper tests + rnaseq snakefile time of under 10 mins.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,7 +6,7 @@ steps:
       - setup.log
   -
     label: wrappers
-    command: source $HOME/anaconda/bin/activate lcdb-wf-test && pytest wrappers/test -v
+    command: source $HOME/anaconda/bin/activate lcdb-wf-test && pytest wrappers/test -v -n6
   -
     label: rnaseq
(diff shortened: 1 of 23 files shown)


daler:
Phew. Working now.

PR #25, opened 2017-04-11
a pile of updates

daler:
This started as adding a first draft of the 4c workflow, but then other things needed fixing. I could go through and cherry-pick into separate PRs, but quite a few things needed fixing so I'm not sure they would all independently pass tests. Hence this monolithic PR. I've described the parts below:

## 4c workflow
Adds `4c.snakefile` for running 4c-ker analysis. This includes `downstream/4c.R`, modules in `lib/fourc`, `config/envs/4c.yaml`, `config/4c-sampletable.tsv`, and `config/test_4c_config.yaml`. I have not generated small example data to use for 4C testing -- this is non-trivial since 4c-ker is really fragile and sensitive to expected 4C coverage patterns. So this workflow remains untested for now.

## Tests
travis-ci has a build timeout of 50 mins. We started hitting that. However travis-ci supports jobs, such that any one job can't take more than 50 mins but multiple jobs can run simultaneously in one build. I've now implemented this in the `.travis.yml`, where there's a `TYPE` env var. For each unique value of `TYPE` an independent job will be triggered with `TYPE` set to that value.. `ci/travis-run.sh` now pays attention to that env var, and runs tests as appropriate. I'm running docs, pytest on wrappers, rnaseq, and references separately. Chipseq and 4c can eventually be run as their own independent jobs. The rnaseq and references are partially redundant, but it buys us a substantial amount of time by running them in parallel rather than sequentially.

It seems like building the environments takes about 10 mins for the rnaseq workflow, which seemed kind of long. Feels like conda is getting slower over time. But I'm not sure there's much we can do about it.

## Docs
Docs live in the `docs` dir as sphinx documentation and are now automatically built and deployed by travis. See the `ci/build-docs.sh` and new entries added to `.travis.yaml`. This works via an encrypted private key whose public key has been added to this repo to effectively give travis-ci push permissions. I had originally built this for bioconda-recipes, added it to lcdblib, and now adding it here.

Upon merging into master, docs will be built and pushed to the `gh-pages` branch accessible at https://lcdb.github.io/lcdb-wf.

## Version bumps

I added a new `ci/dependency_consistency.py` helper script for handling when we need to do version bumps. This finds the set of versions for each tool across requirements.txt and all wrappers, and writes a bash script with sed commands to update everything as needed. Sounds hackish, but I feel like deleting lines from bash scripts is the easiest way to customize rather than exposing some CLI interface.

requirements.txt and all wrappers have been updated to the most recent packages as of Jun 9. I had to fix the bioconda fastq_screen package since the 0.11.1 version was broken.

## rnaseq.snakefile
- Converted pandas `.ix` to `.loc` or `.iloc` as appropriate (since 0.20 now complains HARD).
- added docstrings to rules
- added a bai rule
- added fastq_screen as dependency of multiqc
- added strand-specific bigwig generation. I chose to go with RPKM normalization, since we don't have a good way yet of specifying effective genome size yet.

## other
- added oligomatch wrapper
- added notes about recent atropos wrapper additions
- added `wrappers/test_toy.py`. We can't really do doctests from readmes, but we can do the opposite -- generate readmes from tests. This module is an attempt at this. It works for the simple case; not sure how well it will generalize. I'm adding it because I had written it, forgot about it, rediscovered it and thought it was pretty cool, so hopefully it won't be lost again.


ci/build-docs.sh
```diff
@@ -0,0 +1,111 @@
...
+ENCRYPTED_IV_VAR="encrypted_${ENCRYPTION_LABEL}_iv"
+ENCRYPTED_KEY=${!ENCRYPTED_KEY_VAR}
+ENCRYPTED_IV=${!ENCRYPTED_IV_VAR}
+openssl aes-256-cbc -K $ENCRYPTED_KEY -iv $ENCRYPTED_IV -in $ENCRYPTED_FILE -out key -d
+chmod 600 key
+eval `ssh-agent -s`
+ssh-add key
+
+# clone the branch to tmpdir, clean out contents
+rm -rf $STAGING
```
jfear: Does this need to be wrapped in an if exists statment?

ci/dependency_consistency.py
```diff
@@ -0,0 +1,147 @@
...
+            },
+        }
+    """
+
+    metas = (
+        sp.check_output(
+            ['find', d, '-name', 'environment.yaml'],
+            universal_newlines=True)
+        .splitlines(False)
+    )
```
jfear: This is nice. I would have used os.walk, but this is cleaner.

ci/build-docs.sh
daler: with `-f` there's no error, so this works

4c.snakefile
```diff
@@ -0,0 +1,595 @@
...
+        '4cker-output/{comparison}/cis_k{cis_k}/{sample}_cis_{inter}.bigbed',
+    'cis_treatment_interacting_bigbed':
+        '4cker-output/{comparison}/cis_k{cis_k}/{bait}_{treatment}_cis_highinter.bigbed',
+    'cis_colorized': '4cker-output/{comparison}/cis_k{cis_k}/{bait}_cis_colorized_differential.bigbed',
+    'nearbait_colorized': '4cker-output/{comparison}/nearbait_k{nearbait_k}/{bait}_nearbait_colorized_differential.bigbed',
+}
+
+fills = []
+for comparison, vals in config['4c']['comparisons'].items():
+    bait = vals['bait']
```
daler: Good catch!

wrappers/wrappers/oligomatch/wrapper.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+
+from snakemake.shell import shell
+log = snakemake.log_fmt_shell()
+shell(
+    "oligoMatch "
+    "{snakemake.input.oligos} "
+    "{snakemake.input.sequence} "
+    "{snakemake.output}"
+)
```
daler: Another good catch!

PR #26, opened 2017-04-11
Deeptools bamCoverage

jfear:
Creates a basic wrapper for deeptools bamCoverage.

### DIFF
--- a/wrappers/wrappers/deeptools/bamCoverage/wrapper.py
+++ b/wrappers/wrappers/deeptools/bamCoverage/wrapper.py
@@ -1,6 +1,5 @@
 import os
 from snakemake.shell import shell
-from lcdblib.snakemake import aligners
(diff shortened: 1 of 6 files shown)


wrappers/wrappers/deeptools/bamCoverage/wrapper.py
```diff
@@ -0,0 +1,29 @@
+import os
+from snakemake.shell import shell
+from lcdblib.snakemake import aligners
```
daler: don't think we need this; lcdblib's not in the environment.yaml. (I guess we miss stuff like this during testing when not testing wrappers with --use-conda)

jfear: good catch, that is what I get for copying and pasting.

PR #32, opened 2017-05-12
Add merge

jfear:
Adds a merge function which takes a list of URL's and a list of post-processing functions. Then runs and merges everything.


config/test_config.yaml
```diff
@@ -98,6 +98,34 @@ references:
         indexes:
           - 'hisat2'
           - 'bowtie2'
+
+    ercc:
```
daler: How about something like `r6-11_and_ercc`? Otherwise can't differentiate between other releases.

jfear: good point

PR #35, opened 2017-06-05
Adds another option for atropos PE.

jfear:
I need to be able to dynamically determine if a sample is SE or PE and
then run atropos accordingly.

Problems:
* You cannot use functions as outputs because you need the patterns
* Dynamic outputs gave me problems on large DAGs

I ended up allowing another attribute to be added to `params.R2`,
because params can take functions. Now the wrapper checks for
`snakemake.params.R2` before deciding that `out_R2` should be none. Then
I needed to add a PHONY rule to point to the atropos rule so that if R2
is needed it can be generated. Note with the PHONY rule you need to make the output the final read 2 FASTQ, while in the atropos rule you would need to set read 2 to some temp filename for things to work correctly.

```python
def _input_func_atropos(wildcards):
    """Determine if the sample is PE or SE"""
    flags = some function to pull in se or pe info
    if 'PE' in flags:
        return {'R1': expand(fastqs['r1'], **wildcards)[0], 'R2': expand(fastqs['r2'], **wildcards)[0]}
    else:
        return {'R1': expand(fastqs['r1'], **wildcards)[0]}

def _params_r2_atropos(wildcards):
    """function to make temp R2 if pe."""
    flags = some function to pull in se or pe info
    if 'PE' in flags:
        return expand(patterns['atropos']['r2'], **wildcards)[0] + '.tmp.gz'
    else:
        return None

rule atropos:
    input: unpack(_input_func_atropos)
    output: R1=temp(patterns['atropos']['r1'])
    params: R2=_params_r2_atropos
    threads: 8
    wrapper: wrapper_for('atropos')

rule atropos_phony:
    input: rules.atropos.output
    output: temp(patterns['atropos']['r2'])
    shell: """
    mv {output[0]}.tmp.gz {output[0]}
    """
```

This is kind of a specific case, but maybe it will be generally useful. Using the `dynamic` support in snakemake would be a better solution, but it was taking hours to build the DAG and does not seem to be super stable.


daler:
Fancy! This could come in handy for aligners, prob bwa and star.

Could you alternatively have two rules, one for PE and one for SE, and only include the R2's trimmed fastq in the targets?

PR #36, opened 2017-06-07
channel priority

daler:
We're in the process of fixing a ton of bioconductor and R packages in bioconda. In the meantime, to get tests to pass I'm playing around with channel priorities and may need to pin some versions.


daler:
Good, looks like it was a channel priority. bioconda is moving towards removing the `r` channel, so we'll have to update this again at some point.

PR #38, opened 2017-06-13
refactor into common.py

daler:
An attempt to clean up individual snakefiles' boilerplate by having them import stuff from `lib.common` for setting up the shell, getting sampletables, and reference dirs.

Also improves docstrings.

**edit Jun 21 2017:** turns out bumping version was a huge pain. See https://github.com/lcdb/lcdblib/pull/41 for some details. I also figured out here that we need to pin R in order to avoid a readline error, and that the dupRadar package doesn't specify that it requires KernSmooth which used to be a built-in package but is no longer in the most recent R.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ before_script:
   - export JAVA_HOME=
   - export TMPDIR=/tmp
 script:
-  - travis_wait ci/travis-run.sh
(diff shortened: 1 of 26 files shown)


daler:
Phew. The month-long slog-of-a-PR finally passed . . . .

PR #43, opened 2017-08-18
[WIP] chipseq workflow

daler:
This PR adds a chipseq snakefile. It needed some example data over in `lcdb-test-data`, which is now downloaded with the `ci/get-data.py` script as part of the test setup.

Some points of interest:

- the spp wrapper got surprisingly complicated due to having to fix some problems in spp itself but without patching the source. Since spp is R, and is not really CLI driven, I've exposed a bunch of settings to be controlled by `params:` in the rule. This is accomplished by building a template R script (with some parts being added conditionally) and then filling in the placeholders at the end.

- I usually want to do a peak-calling run using pooled replicates to increase coverage, and use that for downstream analysis. macs2 supports multiple replicates (it pools them internally) but spp does not. The spp wrapper supports multiple bams by merging and re-removing duplicates, sending the result as a single bam to spp.

- there's a new `chipseq: peak_calling:` section of the config to control which samples should be combined together and where per-run settings can be configured. This is way easier to reason about and edit than trying to jam all that info into a sampletable. We might be able to do something like this for RNA-seq (selecting which samples should be run through which models, etc). Not sure yet.

- the chipseq sampletable includes a `label` column to indicate grouping for technical replicates. I don't really like the name of this column, but the functionality works just fine.

TODO:

- [x] deeptools plotFingerprint
- [x] bigwigs for pooled? bigwigs for peak-caller-specific signal output?
~~track hub~~ #47
~~ChIPQC stuff from Jake (Hussain is working on this)~~ #48


daler:
@jfear adding ghostscript as a dep seems to have solved the testing issue

PR #44, opened 2017-08-23
Increment FlyBase to r6.16

jfear:
This PR increments to the latest FlyBase version (r6.16). It also modifies method names for the post-processing steps that appear to have changed.


daler:
Thanks!

jfear:
Strange, none of the conversion tables are being created in the GTF folder.

jfear:
Guess I need to add the annotation hub info from the test config.

PR #45, opened 2017-08-23
Add annotation hub

jfear:
Realized that I needed to add the annotation hub, gene list, and gffutils specific config to the `config.yaml`.


jfear:
Do you think I should go ahead and add these settings to r6-11 too?

daler:
Sure!

jfear:
Maybe it is not just a simple time out issue. Looks like there are some problems rendering the knitr report. I wonder if adding the r-annotationhub is updating another dependency that is breaking knitr. I will test locally and see if I can pin point the issue.

PR #50, opened 2017-08-31
add rnaseq trackhub and config

daler:
(I just mistakenly pushed this directly to master; I then reverted and then changed the repo settings so that master branch is protected)

This will build an RNA-seq track hub of bigwigs, with configuration for coloring, sorting, and organization on the hub config page.

Not sure where it should live yet. `downstream` maybe? This should also work with the chipseq workflow with minor modifications, so maybe it should be generalized to support different kinds of trackhubs? There are also some functions in here that could either be put in [trackhub proper](https://github.com/daler/trackhub) or could be part of a trackhub helpers module.

Edit: oh, and it needs tests too.

### DIFF
--- a/config/hub_config.yml
+++ b/config/hub_config.yml
@@ -1,0 +1,38 @@
+hub:
+  name: 'example-rnaseq'
+  short_label: 'example-rnaseq'
+  long_label: 'example-rnaseq'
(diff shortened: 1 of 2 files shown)


daler:
replaced by #64

PR #51, opened 2017-09-01
Downloads the precomputed FlyBase annotations that map symbol, fbgn, and old annotations.

jfear:
I wanted to download the precomputed file for mapping old FBgns and symbols to current

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -89,9 +89,9 @@ references:
               ahkey: 'AH49581'
               keytype: 'ENSEMBL'
 
-      precompute:
+      fb_annotation:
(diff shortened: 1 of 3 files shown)


daler:
Currently the `fb_annotation` is sort of a top-level key at the same level as `gtf` and `fasta`. I think I'd rather have a `annotation` top-level key that configures the creation of various TSV lookup tables. The existing AnnotationDbi stuff can then be just one way of getting the annotations and these synonyms and annotations can be other ways.

jfear:
I agree, that sounds better. Trying to think the best way to do this. Maybe we can chat about the implementation.

PR #66, opened 2017-10-25
add orig_filename to sampletables

daler:
It's currently undocumented (unless you read the snakefiles) that you can provide an `orig_filename` column to the sampletables and have them automatically symlinked over to sample directories. Super useful for mapping annoying core filenames over to something more meaningful. This PR modifies the example data download to use a different directory structure, and uses the `orig_filename` column in the sampletables to map the files in the new structure over to where they're expected. That way 1) the functionality is more discoverable and 2) it's actually tested now.


jfear:
Awesome. That is super useful and I completely forgot that the functionality was there.

PR #68, opened 2017-11-17
Module-ize patterns and targets

daler:
@jfear this is what we were talking about yesterday . . .

I moved `patterns` and `targets` generation to `lib/patterns_targets.py`; all snakefiles now use that. Patterns are defined in yaml files in `lib/`. So they can be edited, but that's assumed to not be a common use-case. Individual snakefiles now import and create an instance of the respective `ChIPSeqConfig` or `RNASeqConfig` object as `c`, and within the snakefiles I added `c.*` to all references to previously global vars (e.g., `s/patterns/c.patterns/`). Hopefully this also makes it more obvious where the patterns and targets are coming from.

Reorganization: all workflows are moved to their own `workflows/` subdirectory so they can run simultaneously and have independent configs.

Next up will be a `figures.snakefile` that imports the new patterns/targets config objects so that the filenames are easy to get to when accessing subworkflows.

If you would please take a look at this and make any changes/suggestions before things get too crazy, that would help a lot.


daler:
@jfear I just added `figures.snakefile` and it's working with subworkflows to chipseq and rnaseq and reading in the respective config files from those other workflows.

daler:
Thanks, and this was just this morning's work :) 

Renaming to Snakefile is probably a good idea. And I was already thinking about the custom patterns yaml, so I'll add that.

Would you mind taking a stab at the `SeqConfig` inheritance?

daler:
Oh, and yes planning on moving 4C to this same framework, but there are all sorts of problems with the 4C algorithm we're using such that this will be shelved for a while. For now, just assume rnaseq and chipseq.

daler:
(never mind, i added the inheritance here in this next round of commits)

daler:
@jfear when you get a chance can you review again please?

jfear:
Looking now.

lib/patterns_targets.py
```diff
@@ -19,81 +56,129 @@ def update_recursive(d, u):
...
         SeqConfig.__init__(self, config, patterns)
-
-        self.sample_dir = self.configdict.get('sample_dir', 'samples')
-        self.agg_dir = self.configdict.get('aggregation_dir', 'aggregation')
-
-        self.fill = dict(sample=self.samples, sample_dir=self.sample_dir, agg_dir=self.agg_dir)
+        self.sample_dir = self.config.get('sample_dir', 'samples')
+        self.agg_dir = self.config.get('aggregation_dir', 'aggregation')
+        self.fill = dict(sample=self.samples, sample_dir=self.sample_dir,
+                         agg_dir=self.agg_dir)
```
jfear: Would it be better to use the sample table to fill instead of a fill dict? 

While the fill dict is more explicit, using the sample table is more flexible. We could add arbitrary fill patterns to the YAML and fill them with arbitrary values in the sample table.

jfear: Or I guess if we want custom fills, we could just subclass and make modifications we would need.

include/WRAPPER_SLURM
```diff
@@ -30,10 +20,9 @@ if [[ ! -e logs ]]; then mkdir -p logs; fi
     --verbose \
     --cluster 'sbatch {cluster.prefix} --cpus-per-task={threads}  --output=logs/{rule}.o.%j --error=logs/{rule}.e.%j' \
     --use-conda \
-    --configfile config/config.yml \
+    --configfile config/config.yaml \
```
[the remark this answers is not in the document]
daler: Same here, and I go back and forth. I guess in the absence of better criteria, `yaml` is more explicit than `yml`.

lib/patterns_targets.py
```diff
@@ -19,81 +56,129 @@ def update_recursive(d, u):
...
             self.path = config
 
-        configdict, pth = common.resolve_config(config)
-        self.configdict = configdict
+        self.config = common.resolve_config(config)
 
-        common.get_references_dir(config)
+        # Read the config file and extract all sort of useful bits. This mostly
+        # uses the `common` module to handle the details.
+        config['references_dir'] = common.get_references_dir(config)
```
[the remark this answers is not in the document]
daler: Thanks. [Previously](https://github.com/lcdb/lcdb-wf/pull/68/files/30a011b4599a957b7291ff5cab9a8f6857b29ac2..6f588ca1c3f4dff50155ab8214f34f4d3b327eb6#diff-062cede901db6c97c3ac1bf03a3865b5L403) the function had the side-effect of adding to the passed dict. Actually not sure why this passed . . .

lib/patterns_targets.py
daler: For that to happen we'd have to attach `sample_dir` and `agg_dir` as columns to the sampletable, right?

daler:
After using this a bit on an actual analysis, I added another example to the figures snakefile to demonstrate how to grab a reference and using a downstream script.

PR #69, opened 2017-11-21
rnaseq.Rmd fixes for GO/KEGG and AnnotationHub

palmercd:
AnnotationHub: user specifies "Genus species" instead of AnnotationHub ah code. The species name is then pattern matched against the OrgDb species vector and the ah code is determined. User can override by manually specifying ah code. Requires exact match and stops if fails.

GO/KEGG plotting: some ontology terms are long enough to interfere with plotting as apparently y-axis labels take precedence over plot region for ggplot. Automatically truncates terms at 75th percentile of term lengths, adds "..." to end of truncated terms.

GO/KEGG plotting: point legend annotations are automatically set to "experiment{1,2,...}" from clusterProfiler, ignoring res.list specification from user. Modified m.sub to contain experiment codes based on user-specified list keys. Respects user ordering.

### DIFF
--- a/downstream/rnaseq.Rmd
+++ b/downstream/rnaseq.Rmd
@@ -48,11 +48,36 @@ Note: if you're unfamiliar with any of the plots or tables here, see the
 # Try disabling this when running locally to get nicer figures
 # options(bitmapType='cairo')
 
+
+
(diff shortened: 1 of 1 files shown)


daler:
replaced by #70

PR #79, opened 2018-01-17
set latency wait time in wrapper to 300

daler:
like it says on the box


daler:
test failing on multiqc, which is a known issue currently being fixed in bioconda.

PR #83, opened 2018-03-01
update duplicate handling in chipseq snakefile

palmercd:
chipseq snakefile markdups currently uses bowtie2 bam as input, ignoring the output of unique. this is sporadically detectable as inconsistencies between the step 4 and step 5 libsizes. it can result in several million reads being incorrectly included in the markdups output.

the patch changes the markduplicates rule to use the unique output bam as input.


daler:
Tests were failing due to an API change in the latest version of deepTools. That's now fixed in #84, and I added this change over there as well, so closing this one.

PR #84, opened 2018-03-02
Deeptools fixes

daler:
### deeptools arg fixes for new version
New version of deeptools (>3.0.1) has a different way of specifing args for normalization. This PR fixes the bamCoverage commands and pins the deeptools version in requirements to be at least this version. Disabled control bams for plotFingerprint, now using `--smartLabels`, and reduced the number of samples. The net effect is to reduce run time and complexity of the rule.

The deeptools arg change was one of the reasons for the tests failing in #83; since I'm making the fixes here needed for that PR, I'm incorporating #83 here (and closed that one).

### updated test data
Tests were then failing because #83 filters out multimapping reads for chipseq. Turns out in our test data, there are then no reads left in the bam files. I've updated and regenerated the example data; https://github.com/lcdb/lcdb-test-data/compare/fix-multimapping-chipseq shows the updates. This now handles multimapping and uniquely mapping reads explicitly and separately. Those new test data are now used in this PR.

### updated spp wrapper
In the SPP wrapper, there's now a check to remove any chromosomes that have too few reads. Hopefully this should resolve the kind of error we see a lot with spp. And it allows the test data to run.

### Now using Circle-CI
In testing these changes on travis-ci, the builds would often time out. The longest step is actually the conda env creation, but travis-ci doesn't support caching. In bioconda we recently moved from travis-ci to circleci, so I figured I'd give it a shot here.

This actually took the most effort in the PR, but I think it's worth it. Now using caching, it takes about 25 mins for everything to run. I also fixed some fastqc dependency issues.

Major caveat: we are limited to 1500 minutes a month on circleci. I blew through most of that (75%) while getting these tests to run. I have not yet enabled the colocalization, external, or figures workflows. The config for circleci is in `.circleci`; I did not remove any of the travis-ci stuff in case we need to move back to that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ variables:
       run:
         name: chipseq workflow
         command: |
+          export PATH="`pwd`/miniconda/bin:$PATH"
           cd workflows/chipseq
           source activate lcdb-wf-test
           snakemake --configfile config/config.yaml --use-conda -j2 -T -k -p -r
@@ -60,6 +61,7 @@ variables:
       run:
         name: references workflow
         command: |
+          export PATH="`pwd`/miniconda/bin:$PATH"
           cd workflows/references
           source activate lcdb-wf-test
           snakemake --configfile config/config.yaml --use-conda -j2 -T -k -p -r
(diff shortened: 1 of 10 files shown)


daler:
Finally passed!

PR #95, opened 2018-04-04
picard java VM memory reduction

palmercd:
This is a weird one. I tried actually setting the picard java VM ram to 32g and repeatedly got segfaults, which remained if I increased ram, messed with the clusterconfig ram setting, made there be bigger separation between VM memory and node ram, etc.

What ended up working was _reducing_ the VM ram, based on looking at prior RNAseq snakefiles that had worked ok. I guess this is just some issue with incorrect handling of large memory allocations? Whatever it is, it definitely fixes the issue for me. That being said, I would certainly not object to waiting until this is replicated.

### DIFF
--- a/workflows/rnaseq/Snakefile
+++ b/workflows/rnaseq/Snakefile
@@ -407,7 +407,7 @@ rule markduplicates:
     log:
         c.patterns['markduplicates']['bam'] + '.log'
     params:
-        java_args='-Xmx32g'
+        java_args='-Xmx20g'
         # java_args='-Xmx2g'  # [TEST SETTINGS -1]
     shell:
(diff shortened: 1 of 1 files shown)


daler:
I don't think this will *hurt* anything, so might as well merge it and we can revisit this later if needed.

daler:
@isabelleberger could this be what was causing mysterious hang-ups a while ago?

isabelleberger:
Definitely could be... I think the java_args parameters were updated in the chipseq-wf a bit ago and I haven't had any of the hang-up issues since. But could have been something else that was modified in the time before I pulled too.

PR #103, opened 2018-04-12
SICER integration and mappable genome handling

palmercd:
Candidate code for SICER integration into ChIP-seq pipeline. Features:

-SICER runs now supported in ChIP-seq using similar method as with MACS2 and spp. 

-mappable genome proportion or count is now available to peak callers in each run block by injection from the config file.

-SICER and macs2 wrappers both intercept user-specified or genome-default mappable genome information and specify it to the program call.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
- version: 2
+version: 2
 
 variables:
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+ version: 2
 
 variables:
 
@@ -155,5 +155,5 @@ workflows:
           filters:
             branches:
               ignore:
-	        - sicer-integration
+                - sicer-integration
                 - master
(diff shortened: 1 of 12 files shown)


palmercd:
The circleci error is that sicer is not being provided python2.7 as is specified both in the sicer bioconda recipe and in the sicer wrapper environment specification. For whatever reason, it tries to get python2.7 and instead it gets global python3.

The failure of circleci is not replicable elsewhere: not on my local machine, nor on cluster. I think this is some sort of ci-specific problem. Looking around, it seems sometimes miniconda and pyenv in the circleci backend conflict, with different versions of python not being available when you expect them to be. I will not ask for a custom bypass to testing, but for the moment I will leave this branch around and clone it directly to use sicer for active projects.

palmercd:
Current suboptimal status is to encode a conditional in the SICER wrapper that messes with the environment's installation of SICER.sh only when the system is circleci. This forces SICER to use python2.7 from within the environment instead of incorrectly resolving "python" to "python3.*"

PR #117, opened 2018-04-16
Assorted modifications 16-20apr2018

palmercd:
1) assorted SICER bugs from lcdb-126:

   - biowulf rejected the original circleci workaround, so replacing it with a local (tmp) copy of SICER.sh with hardcoded python2 path
   - clusterconfig gets a SICER-specific handler
   - for now, just a comment that SICER emits p==0 in some instances, leading to "inf" entries in the broadPeak output file

2) SICER support in chipseq_trackhub.py from lcdb-125:

   - SICER wrapper now converts the SICER-native wig output files to bigWigs before dumping them in the output directory. Requires use of chromsizes file.
   - chipseq_trackhub.py recognizes SICER-style output and adds the pre- and post-filter signal bigWigs to the chipseq signal view.

3) rnaseq.Rmd generates count data from lcdb-77:

   - Salmon normalized counts are aggregated by gene
   - uses (bad and incomplete) orgdb lookup to map from ENSEMBLTRANS to ENSEMBL
   - per-sample counts are then pasted onto every output DE file
   - a separate counts file (ie all.tsv.counts) is now emitted, one per analysis object. Contains group-combined DESeq2 normalized count data. It's conceptually useful for model checking, to be sure all the various directions of effect make sense. This has not been tested on interaction models but is designed to work on them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,7 +141,7 @@ workflows:
            branches:
              ignore:
                - master
-	       - sicer-integration
+               - sicer-integration
       - rnaseq:
           requires:
             - initial-setup
@@ -149,7 +149,7 @@ workflows:
             branches:
               ignore:
                 - master
-		- sicer-integration
+                - sicer-integration
       - references:
           requires:
             - initial-setup
@@ -157,4 +157,4 @@ workflows:
             branches:
               ignore:
                 - master
-		- sicer-integration
+                - sicer-integration
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,6 +141,7 @@ workflows:
            branches:
              ignore:
                - master
+	       - sicer-integration
       - rnaseq:
           requires:
             - initial-setup
@@ -148,6 +149,7 @@ workflows:
             branches:
               ignore:
                 - master
+		- sicer-integration
       - references:
           requires:
             - initial-setup
@@ -155,3 +157,4 @@ workflows:
             branches:
               ignore:
                 - master
+		- sicer-integration
--- a/workflows/chipseq/chipseq_trackhub.py
+++ b/workflows/chipseq/chipseq_trackhub.py
@@ -16,6 +16,7 @@ import re
 import argparse
 import pandas
 import yaml
+import glob
 from trackhub.helpers import sanitize
 from trackhub import CompositeTrack, ViewTrack, SubGroupDefinition, Track, default_hub
 from trackhub.helpers import filter_composite_from_subgroups, dimensions_from_subgroups, hex2rgb
@@ -73,6 +74,7 @@ subgroups.append(
         name='algorithm', label='algorithm', mapping={
             'macs2': 'macs2',
             'spp': 'spp',
+            'sicer': 'sicer',
             'NA': 'NA',
         }))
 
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
```diff
@@ -134,24 +134,31 @@ workflows:
   test-suite:
     jobs:
       - initial-setup
+         filters:
+	   branches:
+	     ignore:
+	       - sicer-integration
```
daler: I think the tabs here are causing circleci to get confused

workflows/chipseq/chipseq_trackhub.py
```diff
@@ -144,28 +146,30 @@ def decide_color(samplename):
...
-            short_label=sample,
-            long_label=sample,
-            tracktype='bigWig',
-            subgroups=subgroup,
-            source=bigwig,
-            color=decide_color(sample),
-            altColor=decide_color(sample),
-            maxHeightPixels='8:35:100',
-            viewLimits='0:500',
+    if os.path.isfile(bigwig):
```
daler: was this complaining on symlinks? or directories?

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -437,6 +449,48 @@ for (name in names(ll)){
...
+```{r groupcounts, cache=TRUE}
+counts.list <- list()
+
+# compute aggregate (by gene) normalized transcript counts from Salmon
+transcript.counts <- aggregate(transcript.counts, list(transcript.geneids), sum)
+
+# Use normalized counts to get per-level mean counts. Attach Salmon aggregate
+# counts to each output file. This does ~not~ assume any particular model.
+for (name in names(res.list)) {
+    counts.list[[name]] <- res.list[[name]][,c(1:4,7)]
```
daler: would it be safer to use named columns rather than `c(1:4, 7)`? I don't know what these columns are off the top of my head.

wrappers/wrappers/macs2/bdgpeakcall/wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+    '-i {tmp} '
+    '-c {qvalue} '
+    '-l {d_estimate} '
+    '-g {read_length} '
+    '-o {bed} '
+)
+shell(cmds + '{log}')
+
+tmp2 = tempfile.NamedTemporaryFile().name
+# Fix the output file so that it doesn't have negative numbers and so it fits
```
daler: This fixing business ends up being a pretty common thing . . . should we move this to lcdblib so we can do `lcdblib.fix_bedlike(infile, chromsizes, outfile)`?

wrappers/wrappers/macs2/estimate_d/wrapper.py
```diff
@@ -0,0 +1,65 @@
...
+
+d_estimate_file = snakemake.output.d_estimate
+
+# determine if this parameter set has already been estimated for this file
+found_estimate = False
+if os.path.isfile(d_estimate_file):
+    with open(d_estimate_file) as f:
+        for line in f:
+            parsed = line.split()
+            if len(parsed) == 3 and \
```
daler: Are there cases where `d_estimate_file` exists but the estimate is not in that file?

Also, since this `d_estimate_file` is an output for the rule, if it exists when the rule runs but the rule fails for some reason, snakemake will delete it. Is it safe to say that if this rule runs, it should overwrite that file?

.circleci/config.yml
palmercd: yup, it's just a goofy, initially-accidental way to get circle to abandon the test.

workflows/chipseq/chipseq_trackhub.py
palmercd: I honestly don't know what was up with this one, it must have drawn my attention somehow but I recall being tired when I was coding this and cannot remember what the issue was. I apologize and can remove it if you prefer.

workflows/rnaseq/downstream/rnaseq.Rmd
palmercd: yeah so it's grabbing gene, base mean, log2 fold change, stderr, and padj in this situation. It might be safer, it might not. It depends on whether you think it's more likely that the DESeq2 results format itself changes or the colnames change. I think the rest of the file uses colnames? So I can change it to use those as well.

wrappers/wrappers/macs2/bdgpeakcall/wrapper.py
palmercd: I haven't actually observed the out of bounds problem in the wild. But yes it makes sense to move some sort of bed sanitizer to a more public location.

workflows/chipseq/chipseq_trackhub.py
palmercd: Reverted.

workflows/rnaseq/downstream/rnaseq.Rmd
palmercd: Changed to use colnames

PR #121, opened 2018-05-10
allow key to be missing

daler:


daler:
(forcing merge to save on CI quota)

PR #129, opened 2018-07-15
invalidate cache to rebuild env

daler:
Rerunning locally on a new setup I'm not getting the bowtie2 `*.rev.{1,2}` index files. Trying to work out  if that's a new version issue. Re-creating the environment in circleci should test that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ variables:
 
   save_cache: &save_cache
     save_cache:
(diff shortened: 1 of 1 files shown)


daler:
This seems to be a (very) local issue, with the issue only seen on bowtie2 from bioconda in CentOS7 on Xeon E5-2687W v4. Spent the day trying to deconvolve compiler flags, g++ versions, and bioconda infrastructure. No luck, but it's nothing to do with this repo so closing this PR.

PR #138, opened 2018-10-05
adding convert_gtf_to_ucsc

esnaultc:


lib/common.py
```diff
@@ -738,15 +739,58 @@ def convert_gtf_chroms(tmpfiles, outfile, conv_table):
...
-                    if chrom in lookup.keys():
-                        toks[0]= lookup[chrom]
-                        line = '\t'.join(toks)
-                    else:
-                        raise ValueError(
-                            'Chromosome "{chrom}" not found in conversion table '
-                            '"{conv_table}"'
-                            .format(chrom=chrom, conv_table=conv_table)
-                        )
+                    if not line.startswith("#"):
```
daler: Ah yes, good catch on this!

PR #146, opened 2018-11-27
rm -T completely; fix #145

### LINKED ISSUE #145: -T comment in WRAPPER_SLURM breaks script
Leaving the comment line

`     # -T \`

in include/WRAPPER_SLURM causes the snakemake command to be incomplete and leads to various issues depending on the configuration. Most recently the RNA-seq workflow attempted to run but didn't know to use conda environments and attempted to run sequentially, and then eventually crashed ~40% of the way through, on I believe rseqc.

Removing the line entirely resolves the issue.

daler:


daler:
The error is that fastqc, when building the HTML in Java, complains about fonts. This has historically been an issue with fastqc packages (and openjdk which it depends on). I think this is only related to headless systems like circleci that don't have fonts etc installed. Trying a bunch of fixes . . .

PR #167, opened 2019-06-12
Minor fixes

daler:
- automatically detect PE samples and adjust featureCounts arguments appropriately
- fix rnaseq.Rmd so that the KEGG results correctly make it into the enrichment list


daler:
Closing in favor of #170

PR #168, opened 2019-07-16
Update the adapters according to cutadapt docs

njohnso6:
In the documentation here: https://cutadapt.readthedocs.io/en/stable/guide.html#basic-usage
it says Illumina now prefers more specific sequences when cutting the adapters as the previous ones were less specific and present elsewhere in the genome. This is apparently true at least for True-Seq. This is a separate pull request since it is in the actual code and not the test.


daler:
Thanks, I hadn't seen that (for future reference [this is the section](https://cutadapt.readthedocs.io/en/stable/guide.html#illumina-truseq)). They make a good point that the short sequence shows up multiple times in the human genome. Would be interesting to see the impact of those prior recommendations -- could be that some loci have been undersampled for years.

Those updated docs also say that since bwa-mem and STAR do soft-clipping, adapter trimming might not even be needed. By default we're using bowtie2 for ChIP-seq which does not soft-clip by default -- you need the [`--local` flag](http://bowtie-bio.sourceforge.net/bowtie2/manual.shtml#local-alignment-example). HISAT2 [does do soft-clipping](https://ccb.jhu.edu/software/hisat2/manual.shtml#reporting).

Ever since ["An Extensive Evaluation of Read Trimming Effects on Illumina NGS Data Analysis"](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0085024) came out showing that quality trimming was worth it, I've been using quality trimming. It may be time to re-evaluate, as more modern aligners' soft-clipping may remove this need as well. In that case we could consider removing cutadapt altogether.

On the other hand, I see there are quality issues with NextSeq that may not be handled by soft-clipping and that we may want to consider adding as options (e.g. [cutadapt's NextSeq mode](https://cutadapt.readthedocs.io/en/stable/guide.html#quality-trimming-of-reads-using-two-color-chemistry-nextseq)). And if cutadapt is needed for NextSeq, we might want to keep the cutadapt step in after all. It's not likely to hurt anything, just takes up time and space.

Either way, in #170 I had removed the adapters.fa in favor of simplifying the rule, so I've incorporated the changes here in 5dcf003. Keeping this open for now until that PR is merged.

daler:
Closing this in favor of the now-merged #170 which had these changes ported over.

Issue #179, opened 2019-07-22
rnaseq trackhub bigwig nomenclature


daler:
The trackhubs (and bigWig rules) for RNA-seq have the poor nomenclature of "sense" and "antisense". These should be changed.

The way it works now is that bigWigs are created for reads mapping to either strand in the BAM file, but this is independent of whether the assay reports the same or opposite strand reads for a transcript's origin strand. So there is the additional bigwig symlink rule that makes it easy to swap these without having to wait regenerate the bigwigs (which can take a while).

The names of the bigwigs should be chosen carefully to reflect these differences in BAM vs transcript strand.

daler:
After some discussion, we arrived at the following:

- The existing symlink rule to rename bigwigs is not elegant, and really just serves as a way to save time on creating bigwigs when the strandedness of the kit is not known.

- There is no obvious way to name symlinked bigwigs -- pos/neg vs plus/minus vs forward/reverse are all unclear as to which should be the final one.

- No obvious way to signal to the user, via filenames, that all bigwigs should be kept. That is, adding `_final`, to have e.g., `.pos_final.bigwig`, or adding `tmp` or `intermediate` to the original bigwigs might send the wrong message that it's OK to delete. 

- Setting the original bigwigs to read-only would not be consistent with the rest of the workflow.

Fundamentally, the reason for symlinking was to handle cases where the user is not sure about kit strandedness (which, in practice, is pretty often). Re-running the bigwigs has a time and compute cost. However, performing the symlinks, documenting, and managing has a complexity and maintenance cost.

In the end, we decided to:

- remove the symlinking step
- add notes in the Snakefile clearly indicating which params to change
- update rnaseq_patterns and chipseq_patterns appropriately
- update trackhub (rnaseq and chipseq) appropriately

In addition, we proposed a downstream script that is run in order to check for possible issues. One such issue would be that the strandedness is improperly configured. The details of this script still need to be worked out, though.

daler:
Done in #194.

PR #194, opened 2019-07-30
removed bigwig symlinking

hertafeldsr:
This pull request removes the bigwig symlinking step from the workflow. This is in response to our discussion about #179 , during which we decided to scrap the symlinking and make a sort of checkpoint qc step that confirms the strandedness (among other things) before creation of the bigwigs.  The following files were modified:

- 'workflows/rnaseq/Snakefile' (removed the rule 'symlink_bigwigs')
- 'workflows/rnaseq/config/rnaseq_patterns' (removed the 'sense' and 'antisense' entries from the 'bigwig' section
-'workflows/rnaseq/config/config.yaml'  (replaced 'sense' and 'antisense' with 'pos' and 'neg' in the 'merged_bigwigs' section

### DIFF
--- a/test/test_configs/override.yaml
+++ b/test/test_configs/override.yaml
@@ -7,8 +7,8 @@
 #   snakemake --configfile ../../test/override.yaml --config sampletable=/path/to/tsv
 #
 merged_bigwigs:
-  control_sense:
-    sense: []
+  control_pos:
+    pos: []
   treatment_all:
-    sense: []
-    antisense: []
+    pos: []
+    neg: []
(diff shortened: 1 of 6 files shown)


daler:
Thanks, looks like the rnaseq test is failing because the rnaseq_trackhub still has some vestigial `sense` somewhere. The rnaseq-star test is a less obvious failure, I'll dig a little more to see if I can figure out what's going on there.

hertafeldsr:
All fixed now! The override configs in the test folder still had the 'sense' and 'antisense' nomenclature, so updating that seemed to do the trick!

daler:
Thanks, merging!

Issue #191, opened 2019-07-29
Change featureCounts to pooled BAMS


njohnso6:
This is fix for issue #181

daler:
Fixed in #195

Issue #173, opened 2019-07-22
encourage env paths rather than names


daler:
In the docs, should change to suggesting `conda create -p ./env` rather than `conda create -n lcdb-wf`, as the former has been working well across many projects in shared space. Should include some description of why this is a good idea especially on shared systems.

daler:
Fixed in #195

PR #195, opened 2019-07-31
Nick fixes

njohnso6:
This fixes issues #191 and #173 
The merge conflict that would result will be switching "shell" to "run" in the Snakefile


docs/getting-started.rst
```diff
@@ -67,16 +67,20 @@ variable, so that any executables there are found first. It does not affect any
...
+**It is recommended that you create a separate environment directory for each 
+project and include it with the project**.
+That way you can update packages in each project independently of
+any others, and yet the environment will always be close at hand. This is 
+especially good practice in shared space as others can easily find and activate 
+the environment specific to the project.
+Here we use the name "env" for the new environment to clearly denote its contents, 
+but you can use anything. Note that here we specify the channels to use, which 
+include ``bioconda`` which depends on ``conda-forge``, and ``lcdb`` which provides 
+the ``lcdblib`` package used by these workflows.
```
daler: This is much improved!

workflows/rnaseq/config/rnaseq_patterns.yaml
```diff
@@ -11,9 +11,9 @@ libsizes:
   bam:     'data/rnaseq_samples/{sample}/{sample}.cutadapt.bam.libsize'
 fastq_screen: 'data/rnaseq_samples/{sample}/{sample}.cutadapt.screen.txt'
 featurecounts:
-  s0: 'data/rnaseq_samples/{sample}/{sample}.cutadapt.bam.featurecounts.s0.txt'
-  s1: 'data/rnaseq_samples/{sample}/{sample}.cutadapt.bam.featurecounts.s1.txt'
-  s2: 'data/rnaseq_samples/{sample}/{sample}.cutadapt.bam.featurecounts.s2.txt'
+  s0: 'data/rnaseq_samples/combined/cutadapt.bam.featurecounts.s0.txt'
```
daler: There is an existing `rnaseq_aggregation` directory, created for e.g. the multiqc rule (see [this pattern](https://github.com/lcdb/lcdb-wf/blob/master/workflows/rnaseq/config/rnaseq_patterns.yaml#L24)). I think these new aggregated counts would be better living there.

daler:
Thanks. I'm pretty sure tests are currently failing because the rnaseq.Rmd is written to expect the individual featureCounts runs rather than a single file containing all of them, so we'll need to fix that before merging.

workflows/rnaseq/config/rnaseq_patterns.yaml
daler: (fixed in e867fd7)

njohnso6:
Just to make sure it's recorded, I added some further changes to this branch even though not all the tests pass yet.

daler:
This is looking good -- merging into v1.5rc!

PR #207, opened 2019-08-20
More r patching

daler:
Minor fixes in rnaseq.Rmd:

- use lfc.thresh in helpers
- remove some trailing args
- better handling of annotation hub cache


daler:
Phew, it's about time.

PR #212, opened 2019-09-16
improve deploy script

daler:
This fixes the deploy script so that it only ever pays attention to files under version control.

It also adds an optional argument to build the conda env in the destination.


daler:
Closing in favor of #213

PR #229, opened 2019-10-10
fix incorrect sample ID matching between coldata and featurecounts

daler:
This fixes a bug in cases where some samples are substrings of other samples (e.g., `WT_1_1` and `WT_1_10`).

Rather than use the string matching strategy, this expects a function that transforms filenames into sample names. The default is to use the `lcdbwf.samplename` function.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,8 +1,18 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


daler:
closing, see v1.5.3rc

Issue #232, opened 2019-11-02
rnaseq.Rmd should have more 'dependson' in chunks


daler:
RMarkdown code chunks can use `dependson="chunkname"` or `dependson=c('chunk1', 'chunk2')` to build a DAG of dependencies between chunks.

Often we'll delete the cache and re-run when things change/don't change as expected.

We should do a review of the chunk dependencies and add them where needed to minimize the need to delete the cache.

daler:
Closed in #235

Issue #223, opened 2019-10-05
.gitignore improvements


daler:
After a run there are additional files that are created that shouldn't be committed -- AnnotationHub, snakefile logs, various rnaseq/downstream files.

daler:
addressed in #235

PR #234, opened 2019-11-12
.gitignore, rnaseq.rmd dependencies, rseqc read_distribution

njohnso6:
Issues #223 and #232 
The dependency thing should be tested to see if the functional enrichment cache is too easily invalidated. The parent objects it relies on ma y be be changed in previous chunks but the attributes themselves may not be. The documentation is unclear at what level that operates. IT appears it may be conservative at the level of the object. But it is not clear.

### DIFF
--- a/workflows/rnaseq/Snakefile
+++ b/workflows/rnaseq/Snakefile
@@ -760,6 +760,20 @@ rule rseqc_infer_experiment:
     shell:
         'infer_experiment.py -r {input.bed12} -i {input.bam} > {output} &> {log}'
 
+rule rseqc_read_distribution:
+    """
+    read distribution plots
+    """
+    input:
+        bam=c.patterns['bam'],
+        bed12=c.refdict[c.organism][config['gtf']['tag']]['bed12'],
+    output:
+        txt=c.patterns['rseqc']['read_distribution']
+    log:
+        c.patterns['rseqc']['infer_experiment'] + '.log'
+        
+    shell:
+        'read_distribution.py -i {input.bam} -r {input.bed12} > {output} &> {log}'
 
 rule bigwig_neg:
     """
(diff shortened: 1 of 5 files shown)


workflows/rnaseq/Snakefile
```diff
@@ -760,6 +760,20 @@ rule rseqc_infer_experiment:
...
+    """
+    read distribution plots
+    """
+    input:
+        bam=c.patterns['bam'],
+        bed12=c.refdict[c.organism][config['gtf']['tag']]['bed12'],
+    output:
+        txt=c.patterns['rseqc']['read_distribution']
+    log:
+        c.patterns['rseqc']['infer_experiment'] + '.log'
```
daler: `s/infer_experiment/read_distribution/g`

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -29,7 +29,10 @@ output:
 ```
 
 ```{r global_options, include=FALSE}
-knitr::opts_chunk$set(warning=FALSE, message=FALSE)
+knitr::opts_chunk$set(warning=FALSE, message=FALSE, autodep=TRUE)
+
+#automatically identify cache dependencies
+knitr::dep_auto()
```
daler: whoa . . . have you had a chance to test this in practice?

Reading through https://yihui.name/knitr/demo/cache/ to learn more about other things in the cache I've missed, I saw https://github.com/yihui/knitr/issues/238 which uses the md5sum of an input data file. 

What are your thoughts on using the md5sum of the featurecounts file for this?

njohnso6: It takes about .03 seconds for a 5 MB file, not a bad idea. 
Perhaps also the sampletable?
I have not had a chance to test this in practice using our data and variables. However, I have tested it with our code and dummy variables. I'll test it a little now to see how it fares.

daler:
Closing in favor of #241 which is mostly a duplicate of this one

PR #238, opened 2019-11-16
Actions test 2

daler:


daler:
OK, this is not going to work -- there is a cache size limit of 400MB which the conda env blows past. Closing this and starting another branch that attempts to build and then use a docker container.

PR #239, opened 2019-11-16
GitHub action docker

daler:


daler:
Some notes on this:

- caching the conda env is not going to work, since the cache limit is something like 400 MB and this env is a lot bigger
- I've worked out how to build docker images and upload them to the github package repository
- I have not figured out to *use* those uploaded images as part of the "container:" section of a job. You can log in as part of a step (which is a more granular level than a job) but cannot log in at the job level. Yet we need to access the container at the job level.

What I'm trying to do is:

- if requirements.txt changed, build a new docker image and push to github package repository
- use that container for all subsequent steps

daler:
See https://github.com/daler/tmp for a way of building a docker container if requirements.txt (or the Dockerfile) have changed, pushing it to DockerHub, and pulling that down to run a simple test.

It's a bit slower than I would have thought -- takes a while to build the docker image, upload it, and then pull it back down later. And it does not work well for our use-cases where we need to mess with the environment (e.g. add new packages to support new analyses).

So while the proof-of-concept works, I'm not convinced it gains us anything over conda envs.

I would however like to start archiving the built env (`conda pack`) and storing that as an artifact to download for a quick, verified way of creating a conda environment for a new deployment.

PR #247, opened 2020-03-05
Njohnso6 patch 2

njohnso6:
This allows nested.lapply to be run with bitr_override even if all the genes in deg.list do not have a match in the ordb.


daler:
Should be addressed in a34d0d1, closing this

Issue #214, opened 2019-09-21
improvements to STAR rule


daler:
1. Use `--outSAMtype BAM SortedByCoordinate`
2. Use 2-pass mapping. STAR docs recommend doing the first pass on each sample and then providing all identified junctions to each samples' second pass. Not sure if this is worth the complexity of continuing to support multiple aligners.

daler:
This was implemented in b242d354e20738f11172070f5722d37f514083dd.

PR #278, opened 2020-05-07
implement STAR 2-pass alignment [WIP]

daler:
This addresses #214, improving the STAR step.

This optionally implements the method recommended in the STAR docs, doing a 2-pass alignment. The STAR index is built using the GTF file (this was already in master). The first pass throws away the alignments but keeps the splice junction database. We do this this for all samples. Then the second pass realigns each sample separately, but each time feeds *all* of the splice junction databases together (for improved novel splice junction detection, according to the docs) and creates the final BAM.

Because the 2-pass mapping step does actually require aligning twice, this is extra overhead that may not be worth it all the time. So we retain the original 1-pass when `aligner: star` is configured and use the 2-pass when `aligner: star-twopass` is configured.

STAR tends to clutter the directory with log files, so these new rules do a bit of housekeeping. Still need to work out:

- ~~exactly which logs multiqc is parsing -- does pass2 override pass1?~~ *update: pass1 seemed to be winning; excluded it in multiqc config in b583d67*
- ~~consider cleanup of of the SJ.out.tab for pass 1 and pass 2~~ *update: yes, removed in bcce5fa* 
- ~~consider cleanup of the _STARgenome directory for pass 2~~ *update: yes, removed in fa51464*


daler:
Hmm, looks like multiqc is only paying attention to logs from the first pass. Not clear how to handle this. Keeping both sets of logs is useful to evaluate if the 2-pass approach was enough of an improvement with respect to splice junctions, but probably only the second should show up in the multiqc report. I think the right solution is to exclude it in the multiqc config.

daler:
Regarding the SJ.out.tab files, from [this forum answer](https://groups.google.com/d/msg/rna-star/XN0cWBxVFcM/4yoJyQZeAQAJ):

> the SJ.out.tab contains filtered splice junctions detected in the mapping - i.e. if you extract junctions from the BAM and filter them according to --outSJfilter* filters, you would get the SJ.out.tab file.

So we don't gain any additional information by keeping it. bcce5fa sets them as `temp()` files so that they're cleaned up automatically.

PR #283, opened 2020-05-11
Added contrast examples and doc

mitraak:
Added some documentation of different ways of specifying contrasts and examples in the comments section of the results chunk


daler:
This is really helpful information, esp. the part about "intercept-less" models. However I'm thinking about needing to scroll through this every time we're working on the Rmd.

What if we included something on each line that could be grepped out? Maybe lines with `##` could be for these sorts of "extra comments" that are helpful in the general case, esp for newer users, but that could be easily grepped out? 

Come to think of it, we could probably do that for a lot of the other explanatory comments which could clean things up a bit if the user so chooses.

Or is there a better place to put this sort of information? For example we could write and host pages on https://lcdb.github.io/lcdb-wf/ (source [here](https://github.com/lcdb/lcdb-wf/tree/master/docs)). Then the Rmd comments can just become links to those documentation pages. We could even have a section in the docs about interaction terms and include figures.

mitraak:
The part about the 'intercept-less' model can definitely be moved somewhere else, say `help_docs.Rmd` (for collaborators) or the github.io pages (for external).

daler:
Is the other text from `?DESeq2::results`? Or, close enough to it that we can just cite that?

mitraak:
It's similar to DESeq2, so we can cite it. It sort of a TL;DR of that

daler:
@mitraak I moved your comments over to the docs. From the build-docs test on CircleCI, you can click the "artifacts" tab and download the built docs as a tarball; here's the latest at the time I'm writing this: https://4471-74318002-gh.circle-artifacts.com/0/tmp/docs.tar.gz.

The idea is that we can add extensive documentation there and not clutter up the Rmd.

Issue #267, opened 2020-04-30
make FDR and fold-change cutoffs uniform


mitraak:
- make parameter names referring to FDR and FC cutoffs uniform in the plots chunk. This touches the functions:
    - `summarize.res.list`
    - `my.summary`
    - `plotMA.label`

PR #289, opened 2020-05-19
unify FDR and log2FoldChange thresholds

esnaultc:
Unify how FDR and LFC are configured and passed within the RNAseq downstream Rmd.

Edit: addresses #267


workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -395,32 +403,48 @@ res.list <- list()
...
+                  contrast=c('group', 'treatment', 'control'),
+                  alpha=metadata(res.all)[['alpha']],
+                  res=res.all,
+                  type='normal'
+                  ),
     dds=dds,
-    label='Using a log2FoldChange threshold of 0'
+    label='Using a log2FoldChange threshold of 0',
+    alpha=metadata(res.all)[['alpha']],
+    lfc.thresh=metadata(res.all)[['lfcThreshold']]
```
daler, line 422: So we can get the alpha and lfcThreshold from the results object using `metadata()`. . . I think rather than have it here, it would be cleaner to instead pull out this info from inside any helper functions that operate on res.list or on a results object. I mean, that complexity to extract alpha and lfcThreshold does have to exist somewhere; I'd rather not have it here where the user needs to type it in each time.

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -37,6 +37,14 @@ knitr::opts_chunk$set(
 )
 ```
 
+```{r cutoffs, include=FALSE, cache=TRUE}
```
daler, line 40: I'm unsure about having a global alpha and lfc threshold that we use for everything, as we sometimes run two versions (say, one with lfc.thresh=0 and another with lfc.thresh=1). @mitraak could you weigh in on this?

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -478,10 +498,12 @@ for (name in names(res.list)){
   dds.i <- res.list[[name]][['dds']]
   res.i <- res.list[[name]][['res']]
   label <- res.list[[name]][['label']]
+  alpha.i <- res.list[[name]][['alpha']]
+  lfc.thresh.i <- res.list[[name]][['lfc.thresh']]
```
daler, line 502: if alpha isn't entered into the res.list pseudo-manually (see above comment), then here you can use  e.g.,`metadata(res.list[[name]][['res']])[['alpha']]` to figure it out

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -478,10 +498,12 @@ for (name in names(res.list)){
...
 
   mdcat('## ', label, ' {.tabset}')
   mdcat('### Summary of results')
-  print(knitr::kable(my.summary(res.i, dds.i)))
+  print(knitr::kable(my.summary(res.i, dds.i, alpha.i, lfc.thresh=lfc.thresh.i)))
```
daler, line 506: following up on the above comments, `my.summary` can figure out the alpha and lfc thresh on its own which would simplify this

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -496,14 +518,19 @@ for (name in names(res.list)){
   genes.to.label <- res.i[1:10, 'symbol']
 
   mdcat('### M-A plot')
-  print(plotMA.label(res.i, genes.to.label=genes.to.label, col='symbol'))
+  print(plotMA.label(res.i,
+                     fdr.thres=alpha.i,
+                     fc.thres=lfc.thresh.i,
+                     genes.to.label=genes.to.label,
+                     col='symbol')
+  )
```
daler, line 526: ...and maybe plotMA.label can figure it out itself, too?

workflows/rnaseq/downstream/rnaseq.Rmd
mitraak: I have been leaning towards this lately. Multiple lfc thresholds are fine during the exploratory phase, but eventually we usually choose one unless the collaborators want that tweaked.

mitraak: I like the idea of using the `metadata()` function to unify the lfc and fdr thresholds in the helper functions. That will definitely clean up the code.

workflows/rnaseq/downstream/rnaseq.Rmd
mitraak: I agree with Ryan here. I like the idea of using the `metadata()` function to unify the lfc and fdr thresholds in the helper functions. That will definitely clean up the code and we won't need to set it explicitly here.

workflows/rnaseq/downstream/rnaseq.Rmd
mitraak: I think I like the generality of the function. It just needs a data.frame to make the plot. We can expand it to also support a `DESeqResults` object, so something like a tryCatch with `metadata()` and use the alpha and lfc cutoffs if any. If not, just keep the present form.

daler: So something like this?

- the first arg of plot.MA could be a dataframe or a results object
- If dataframe, fdr and fc thresholds must be supplied if they are to be used
- If results object, then use `metadata()` to get the thresholds
- If both results object and threshold(s) are provided...which one wins?

mitraak: yes, exactly. 

I think the last case is the most tricky, and will involve some decisions. I'm thinking explicitly-specified thresholds win? We can output a warning that there is a mismatch

daler: explicit-wins sounds good to me, and a warning would be a nice touch.

mitraak: this also reminds me to update the function with my newer version with uses dplyr piping and cleans it up a lot. I'll add an issue with both these things

workflows/rnaseq/downstream/rnaseq.Rmd
esnaultc: `lfcShrink` does not carry over the res lfcThreshold. Would it make more sense to do the `results()` without threshold and apply an lfcThreshold to `lfcShrink`? Or to apply the lfcThreshold to both `results()` and `lfcShrink`?

daler: Oh no!

It might take writing it out and seeing it both ways to figure out what the best long-term solution will be. I think the major guiding principle is simplicity on the rnaseq.Rmd side (to avoid bugs). If it doesn't become clear after trying it different ways, I think having a separate PR ("dueling pull requests") would be a good way for us to jointly evaluate the options.

esnaultc: The lfcThreshold in `lfcShrink()` is for s-values. Michael Love recommends setting the threshold in `results()` to work with an FDR bounded set ([here](https://support.bioconductor.org/p/110307/)), but that means we can't use the metadata from the lfcShunk res.list to figure out the lfcThreshold.

PR #290, opened 2020-05-21
Degpatterns patch

mitraak:
- fixed `degPlotCluster` issues with 0 size data frame
- also added combined cluster membership TSV

### DIFF
--- a/workflows/rnaseq/downstream/gene-patterns.Rmd
+++ b/workflows/rnaseq/downstream/gene-patterns.Rmd
@@ -1,4 +1,4 @@
-# Gene patterns {.tabset}
+# Gene patterns
 
 We can roughly group genes into expression patterns. This uses the [DEGreport
 package](https://www.bioconductor.org/packages/release/bioc/html/DEGreport.html),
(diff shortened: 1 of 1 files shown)


workflows/rnaseq/downstream/gene-patterns.Rmd
```diff
@@ -108,41 +110,71 @@ for (name in names(ll)){
...
+```{r plot_degpatterns, cache=TRUE, fig.width=12, results='asis', dependson='final_clusters'}
+# unique list of genes in clusters
+uniq.genes <- unique(unlist(lapply(dp.list, function(x) x[['genes']])))
+
+# data frame to hold cluster memberships
+# NOTE: here we assume that all genes are present in any 'res' object.
+#       - this should be replaced by the geneid -> symbol mapping df
+cluster.df <- data.frame(row.names = uniq.genes,
+                         genes = uniq.genes,
+                         symbol = res.list[[1]][['res']]$symbol[uniq.genes])
```
daler, line 130: I think it's a reasonable assumption that the first res has all genes. But should we hold off on merging this until that mapping df is implemented?

mitraak: yes, agreed. this is somewhat of a placeholder

Issue #275, opened 2020-05-04
proposal for streamlining creation of dds objects


daler:
We typically make a handful of dds objects to look at different aspects of the experiment. This ends up creating a fair amount of code duplication, with the corresponding introduction of possible errors.

For example, it's currently pretty messy to remove outliers, especially in cases where the technical replicates need to be merged and ensembl versions need to be stripped. That is, these three things happen:

```r
#1.
dds <- DESeqDataSetFromCombinedFeatureCounts(
  '../data/rnaseq_aggregation/featurecounts.txt',
  sampletable=colData,
  design=~group)

#2.
dds <- collapseReplicates(dds, dds$biorep)

#3.
if (strip.dotted.version){
  rownames(dds) <- sapply(strsplit(rownames(dds), '.', fixed=TRUE),
                                  function (x) x[1])
}
```                                        

When creating different dds objects, the above three steps need to happen each time and it's easy to get mixed up with the variable names. I think one way to streamline this would be to have DESeqDataSetFromCombinedFeatureCounts do the collapsing (if needed) and version strip (if needed) within that function.

The result of this example code is a list of dds objects where they all pull from the same salmon input files and use the same technical replicate collapsing and the same Ensembl version removal.

```r
lst <- list(
  main=list(colData, ~group),
  no.rep.2=list(colData %>% filter(replicate!=2), ~group),
  interaction=list(colData %>% filter(replicate!=2), ~genotype + time + genotype:time)
)

dds.list <- make.dds.list(
   lst, salmon.files=salmon.files, combine.by='biorep', remove.version=TRUE)
```

In this way, `make.dds.list` only needs to know the salmon files, collapse info, and version removal info once. The only things that change are the colData and model.

So . . . just have to write `make.dds.list` to make that happen!

An additional side effect is that the res.list object will no longer store copies of each dds object, we can just refer by list name to the new dds.list object.

Note: this will use the salmon importing through tximport, which has arguments for at least removing ensembl versions, so that can be passed through to the tximport function.

njohnso6:
Benchmarking, if done correctly, revelas lscratch, normal, and shm to have similar times:
Normal? Scratch allocated but not assigned:
user        sys        elapsed
55.523    3.711    62.643
56.57    1.37        57.956

Lscrath officially allocated 
user        sys        elapsed   
59.886    3.770    64.042
52.935    1.527    54.465

With shm
57.783    3.90    62.232
50.792    1.278    52.076

[...]

PR #291, opened 2020-05-24
streamline the handling of dds objects (make.dds.list())

njohnso6:
This pull request creates a helper function, make.dds.list, that can take a list of dds argument lists and combine them into dds objects. The function accepts arguments to DESeq or the DESeqDataSet constructors. It can use either feature counts or transcript counts (salmon). One caveat is salmon files, at this point in time, can only be used if all salmon files will be used (if the sample table is not altered in any way, or the samples filtered). A new issue must be created to address that. 
This pull request also modifies the rnaseq.Rmd to include a template ready for make.dds.list. It modifies the workflow to use the resulting list, and to use it efficiently. res.list no longer holds a copy of each dds object, but simply a label indicating which dds object in the list was used to create it. 

edit: addresses #275


workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -29,7 +30,7 @@ get.orgdb <- function(species, cache, annotation_key_override=NA){
     }
 
     ah <- AnnotationHub(hub=getAnnotationHubOption('URL'),
-             cache='../../../include/AnnotationHubCache',
+             cache=cache_choice,
```
daler, line 33: thanks, I hadn't realized until now that was hard coded!

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -357,6 +358,64 @@ DESeqDataSetFromSalmon <- function (sampleTable, directory='.', design,
...
+        subset.counts <- FALSE
+        colData <- pluck(design_data, 1)
+        design <- pluck(design_data, 2)
+        # arg list passes args to either the CombinedFeatureCounts or FromSalmon constructor
+        arg_list <- pluck(design_data, 3)
+
+        if(is.null(salmon.files)){
+            # Load gene-level counts
+           dds<- exec(DESeqDataSetFromCombinedFeatureCounts,
+                '../data/rnaseq_aggregation/featurecounts.txt',
```
daler, line 389: this should be parameterized

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -357,6 +358,64 @@ DESeqDataSetFromSalmon <- function (sampleTable, directory='.', design,
...
+                sampletable=colData,
+                # NOTE: What design to use?------------------------------------------------
+                #   This object will be used for EDA, so it's recommended to use a single
+                #   factor that describes the samples. We'll be recreating another dds
+                #   later for the actual contrasts of interest.
```
daler, line 394: this comment doesn't hold true (looks like it was copied from the setup for the "initial" dds used for clustering and PCA)

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -357,6 +358,64 @@ DESeqDataSetFromSalmon <- function (sampleTable, directory='.', design,
...
+#' a list of DESeq object. 
+#'
+#' @param deseq_obj_list A named list of lists. Each list should have two to three elements:
+#'       1 should be a sampletable, 2 should be a design. The 3rd is an optional list of arguments
+#'       to pass to the DESeqDataSet constructor.
+#' @param salmon.files txi object from tximport if transcript-level counts desired; NULL forces featurecounts; 
+#'        shared by all dds.
+#' @param combine.by The column to collapse technical replicates by.
+#' @param remove.version Boolean to decide whether to remove version information 
+#' @return A list of dds objects.
```
daler, line 373: could use some more detailed docs here, I'll take a stab at it an push to this PR

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -357,6 +358,64 @@ DESeqDataSetFromSalmon <- function (sampleTable, directory='.', design,
...
+#' 
+make.dds.list <- function(deseq_obj_list, salmon.files=NULL, 
+                          combine.by=FALSE, remove.version=TRUE, ...){
+    make.dds <- function(design_data, salmon.files, combine.by, remove.version, ...)
+    {
+        
+        subset.counts <- FALSE
+        colData <- pluck(design_data, 1)
+        design <- pluck(design_data, 2)
```
daler, line 382: I didn't know about pluck() before, very useful!

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -357,6 +358,64 @@ DESeqDataSetFromSalmon <- function (sampleTable, directory='.', design,
...
+            # Load gene-level counts
+           dds<- exec(DESeqDataSetFromCombinedFeatureCounts,
+                '../data/rnaseq_aggregation/featurecounts.txt',
+                sampletable=colData,
+                # NOTE: What design to use?------------------------------------------------
+                #   This object will be used for EDA, so it's recommended to use a single
+                #   factor that describes the samples. We'll be recreating another dds
+                #   later for the actual contrasts of interest.
+                design=design,
+                !!!arg_list)
```
daler, line 396: I've never seen `!!!` before -- it looks to be an equivalent of Python's `**kwargs`, is that right?

workflows/rnaseq/downstream/helpers.Rmd
```diff
@@ -593,14 +652,19 @@ attach.info <- function(res, keytype='ENSEMBL', columns=c('SYMBOL', 'UNIPROT', '
 #'
 #' @param res.list Named list of lists, where each sublist contains the following
 #'                 names: c('res', 'dds', 'label'). "res" is a DESeqResults object,
-#'                 "dds" is a DESeq object, and "label" is a nicer-looking
-#'                 label to use.
+#'                 "dds" is either the indexing label for the dds.list object or
+#'                  the DESeq object, and "label" is a nicer-looking
+#'                 label to use. NOTE: backwards compatibility with older versions
+#'                  of lcdb-wf depends on no dds.list object being passed.
```
daler, line 658: I very much appreciate handling the backwards compatibility, but I think we're making enough changes to rnaseq.Rmd that backwards-compatibility will break in many places. I could go either way, keeping this as-is or simplifying to always require dds.list.

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -154,17 +154,15 @@ Here is the sample table with metadata used for this analysis:
 knitr::kable(colData[, colnames(colData)[!colnames(colData) %in% exclude.for.printing]])
 ```
 
-```{r salmon}
+```{r salmon, eval=FALSE}
```
daler, line 157: if the default below is to use `salmon.files=txi` then this should be `eval=TRUE`, right?

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -315,32 +313,27 @@ parallel <- FALSE
...
-dds <- DESeq(dds,
-    # NOTE: betaPrior----------------------------------------------------------
-    #    betaPrior=FALSE is the default in DESeq2 >1.16, but here we set it
-    #    explicitly for consistency across different versions.
-    betaPrior=FALSE,
-    parallel=parallel
-    )
+lst <- list(
+            main=list(colData, ~group),
+            r = list(colData %>% filter(samplename!='sample4'), ~group, list(subset.counts=TRUE))
```
daler, line 334: to make it clearer:

```suggestion
            no_rep4 = list(colData %>% filter(samplename!='sample4'), ~group, list(subset.counts=TRUE))
```

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -453,7 +446,8 @@ help](#Help) section for details.
 
 ```{r, results='asis'}
 # Summarize all experiments
-knitr::kable(summarize.res.list(res.list, alpha, lfc.thresh))
+# NOTE: for backwards compatibility with previous versions of lcdb-wf, remove the dds.list argument.
```
daler, line 449: same comment as above re: backwards compatibility

workflows/rnaseq/downstream/helpers.Rmd
njohnso6: It is very similar. But there are some differences related to named and unnamed parameters. Also, in this case, the following function doesn't have tidy dots, so the exec is also required so the !!! evaluates first. Hadley once again has our back on this [here](https://adv-r.hadley.nz/quasiquotation.html). I actually wanted to put together a few resources I found on the subject. I'll put it in a wiki.

workflows/rnaseq/downstream/helpers.Rmd
njohnso6: a file argument is now allowed as well with the above set to default.

workflows/rnaseq/downstream/rnaseq.Rmd
njohnso6: Not sure if I remembered correctly but set salmon.files=NULL in the example call instead.

daler: Yep, that would do it. So let's keep this PR using featureCounts, and once other PRs are in we'll switch over to using salmon.

daler:
@njohnso6 I did a round of cleanup and changes here, including:

- resolving all the merge conflicts due to recent v1.6rc changes
- requiring the `design_data` to be a named list (I think this makes it more explicit, and is easier to follow given that it mimics what the `DESeqFromXYZ...` constructors look like)
- split out the `make.dds` function into its own first-class function, and put the documentation there and updated it
- added a new section to the `rnaseq.Rmd` docs
- `make.dds.list` is now possibly not needed, as we just need to make the `map()` call in the main `rnaseq.Rmd`

PR #294, opened 2020-06-03
Issue 233 - parallelize functional enrichment

mitraak:
code to parallelize functional enrichment. Will be pushing incremental commits as the code takes shape


mitraak:
Some notes:
- for the test data, the `bplapply` is actually slower than `lapply`, possibly because of having to load the `annotationHub`
    - for 4 workers
       - list size = 2, parallel: 90s, non-parallel: 25s
       - list size = 6, parallel: 115s, non-parallel: 45s
       - list size = 10, parallel: 130s, non-parallel: 65s
- not sure if size of gene list will actually make the parallel operation faster

mitraak:
- testing for gene length:
    - length of gene list in test data: ~10-15
- for list length of length 10 x 2
    - gene list length = 100, parallel: 130s, non-parallel: 75s
    - gene list length = 150, parallel: 135s, non-parallel: 80s

is it worth to parallelize? Or can we just use non-parallel `lapply`?

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -98,11 +99,28 @@ if(!exists(hub.cache)){
...
+orgdb <- ah[[species_accession]]
+```
+
+```{r parallel, cache=TRUE}
+# NOTE: Run DESeq2 on multiple cores.------------------------------------------
+#    By default, we do not run in parallel, however this can be very useful in
+#    experiments with many samples and complex designs.  To run in parallel,
+#    set `nworkers` to the number of available cores and `parallel <- TRUE`.
+#    Then, add the argument `parallel=TRUE` whenever you call DESeq()
+nworkers <- 2
```
daler, line 121: (double check comments and what this default should be)

daler:
Closing, as the functional enrichment has ben refactored to be more general; see #362 and subsequent PRs

PR #298, opened 2020-06-08
sampletable non-unique fastqs

esnaultc:


daler:
I could imagine multiple "pre-flight checks" that we could do, and this would be one of them. Also we should have this for ChIP-seq as well. Can you please do the following: 

- add a new `lib.helpers.preflight()` function that accepts the config object (it should then figure out the sampletable from that and import as a dataframe)
- move this unique fastq file check so that it is within that new function
- have both rnaseq and chipseq workflows call the new preflight() function.

daler:
Looks good, thanks!

PR #320, opened 2020-10-18
DAG simplification

daler:
The ChIP-seq and RNA-seq workflows had many intermediate files as inputs to the `targets` rule. This removes those and only includes the multiqc and the bigwig/bigbed creation.

I'm not sure if this is the right thing to do. It definitely simplifies the DAG. But users just starting out may get confused as to why, after adding a sample, nothing needs to be re-run. That is, before this PR adding a sample to the sampletable would add a BAM to the inputs of the `targets` rule, and would sort of force the rest of the DAG for that sample. But after this PR, if multiqc.html already exists then no jobs would be run by default. Instead they'd need to force-run the multiqc rule to trigger the inputs and therefore the rest of the DAG.

### DIFF
--- a/workflows/chipseq/Snakefile
+++ b/workflows/chipseq/Snakefile
@@ -64,20 +64,9 @@ helpers.preflight(config)
 
 # See "patterns and targets" in the documentation for what's going on here.
 final_targets = utils.flatten((
-    c.targets['bam'],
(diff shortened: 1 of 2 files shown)


daler:
Closing . . . I think I'd prefer pay the cost of slightly messier DAG to have new lines in the sampletable automatically trigger everything.

PR #321, opened 2020-11-25
[WIP] Update index.rst

evajason:
Made a few changes for clarity.


docs/faqs.rst
```diff
@@ -89,9 +89,9 @@ both counts tables (see :ref:`rnaseqrmd` for more).
 Samples need to use the same parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There is no mechanism for specifying sample-specific parameters (for example,
-to use cutadapt to trim some 5' bases from some samples but leave other samples
-alone). Samples that need to treated differently should be split off into
+There is no mechanism for specifying sample-specific parameters. For example,
+to use cutadapt to trim some 5' bases from some samples. Other samples are left
+alone. Samples that need to treated differently should be split off into
```
daler, line 94:
```suggestion
There is no mechanism for specifying sample-specific parameters. For example,
to use cutadapt to trim some 5' bases from some samples but leave other samples alone.
Samples that need to treated differently should be split off into
```

docs/getting-started.rst
```diff
@@ -46,13 +46,13 @@ project to use lcdb-wf means copying over the files you need. There is
 a ``deploy.py`` script that does this for you, but you need to have a copy of
 the repository to get this script in the first place.
 
-The reason to use ``deploy.py`` is that there are many additional files (like
-these docs) and testing files that are not necessarily needed for an actual
-project, so you end up with a cleaner project directory.
+The reason to use ``deploy.py`` is so you end up with a cleaner project directory.
+There are many additional files (like these docs) and esting files that are not
```
daler, line 50:
```suggestion
There are many additional files (like these docs) and testing files that are not
```

daler (changes requested):
Thanks! Minor suggestions below.

evajason:
The list in the [ChIP-seq workflow](https://github.com/lcdb/lcdb-wf/blame/eva-edits/docs/chipseq.rst#L9-L29)does not match the tense of the sentence before. Maybe change "does" to "will do".

evajason:
I think the example [here](https://github.com/lcdb/lcdb-wf/blame/master/docs/faqs.rst#L92-L94) shouldn't be in parenthesis. Seems like important information.

evajason:
missing [list] (https://github.com/lcdb/lcdb-wf/blame/master/docs/config-yaml.rst#L532) of other peak callers in config-yaml.rst

evajason:
Give examples for ["custom parameters"] (https://github.com/lcdb/lcdb-wf/blame/master/docs/faqs.rst#L100-L101). Are these multiple IP and input labels? I was looking in the (config-yaml.rst)[https://lcdb.github.io/lcdb-wf/config-yaml.html#config-yaml] and it wasn't clear to me what "custom parameters" refers. But, I have also never used the ChIP-seq workflow so that might also be why it isn't clear to me.

evajason:
[This](https://github.com/lcdb/lcdb-wf/blame/eva-edits/docs/patterns-targets.rst#L24-L28) could be clearer. I would swap around the order of the sentence:

The metadata configured in the sample table and config file fills in those patterns to create the **targets**.

Because the patterns already exist, the user fills in the metadata, and then the targets are created, right?

evajason:
In [patterns-targets.rst](https://github.com/lcdb/lcdb-wf/blob/eva-edits/docs/patterns-targets.rst#L24-L28) it has both "sample table" and "sampletable" written. It should probably be "sample table" unless directly referring to the sampletable.tsv within the workflow.

evajason:
I think the [Unified approach to references](https://github.com/lcdb/lcdb-wf/blame/eva-edits/docs/index.rst#L38-L46) section is confusing and repetitive. I suggest switching the order to explain the reference wf first. No need to repeat that it's made for customization because that's the first thing on this page. 

Unified approach to references

This system is designed to include arbitrary genomes whether local or on the web. The references workflow is shared by RNA-seq and ChIP-seq and is driven by a config file that specifies URLs for FASTA and GTF files. Set it up once for a website to get lots of genomes you can use for running fastq_screen, and easily include arbitrary other genomes. They can then be automatically included in RNA-seq and ChIP-seq workflows. The references_dir is used as a centralized repository that can be included with all other future projects.

PR #330, opened 2021-02-23
Edits for clarity

evajason:
These are the edits for the comments mentioned in #321. I didn't use that branch because the branch has some issues.


daler (commented):
Looks great!

For those items in #321 that are not resolved here, for example the chipseq peak callers, can you please consolidate those comments in the description of this PR and then close #321? I'll take it from there.

evajason:
Unresolved issues from #321:

- Give examples for [custom parameters](https://github.com/lcdb/lcdb-wf/blame/master/docs/faqs.rst#L100-L101). Are these multiple IP and input labels? I was reading about the config and it wasn't clear to me what "custom parameters" refers. But, I have never used the ChIP-seq workflow so that might also be why it isn't clear to me.

- missing [list](https://github.com/lcdb/lcdb-wf/blame/master/docs/config-yaml.rst#L532) of other peak callers in config-yaml.rst

daler:
Thanks Eva!

PR #333, opened 2021-03-12
Make rnaseq contrast chunks independent.

njohnso6:
Addresses issue #322.

### DIFF
--- a/workflows/rnaseq/downstream/rnaseq.Rmd
+++ b/workflows/rnaseq/downstream/rnaseq.Rmd
@@ -256,19 +256,21 @@ dds.list <- map(lst, lcdbwf::make.dds, salmon.files=NULL, remove.version=TRUE,
 ```{r results, dependson='dds_list', cache=TRUE}
 # Docs: https://lcdb.github.io/lcdb-wf/rnaseq-rmd.html#results
 
-res.list <- list()
-
 # NOTE: Example contrast #1----------------------------------------------------
 #   Using the example data, this compares treatment group to control group.
 #   Change to reflect your experiment.
 # IMPORTANT: Control must always be last
-res.list[['main']] <- list(
+#   Change the letter prefix of the contrast to adjust order of appearance in the html
+#   Each contrast will appear in the res.list without the prefix "contr_[A-z]_".
+#   Make a separate chunk for each contrast.
+contr_a_main <- list(
     res=lfcShrink(dds.list[['main']], contrast=c('group', 'treatment', 'control'), type='normal'),
     dds='main',
     label='Using a log2FoldChange threshold of 0'
 )
+```
 
-
+```{r results2, dependson='dds_list', cache=TRUE}
 # NOTE: Example contrast #2----------------------------------------------------
 #    Using the example data, this compares treatment group to control group but
 #    requiring genes to have >4-fold differences (log2(4) = 2). Change to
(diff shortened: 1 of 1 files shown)


workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -256,19 +256,21 @@ dds.list <- map(lst, lcdbwf::make.dds, salmon.files=NULL, remove.version=TRUE,
...
 # Docs: https://lcdb.github.io/lcdb-wf/rnaseq-rmd.html#results
 
-res.list <- list()
-
 # NOTE: Example contrast #1----------------------------------------------------
 #   Using the example data, this compares treatment group to control group.
 #   Change to reflect your experiment.
 # IMPORTANT: Control must always be last
-res.list[['main']] <- list(
+#   Change the letter prefix of the contrast to adjust order of appearance in the html
```
daler, line 263: This should be added to docs at https://github.com/lcdb/lcdb-wf/blob/master/docs/rnaseq-rmd.rst, and removed here.

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -256,19 +256,21 @@ dds.list <- map(lst, lcdbwf::make.dds, salmon.files=NULL, remove.version=TRUE,
...
+#   Make a separate chunk for each contrast.
+contr_a_main <- list(
     res=lfcShrink(dds.list[['main']], contrast=c('group', 'treatment', 'control'), type='normal'),
     dds='main',
     label='Using a log2FoldChange threshold of 0'
 )
+```
 
-
+```{r results2, dependson='dds_list', cache=TRUE}
```
daler, line 273: the naming of every chunk will probably break the tests that ensure each named chunk has a corresponding entry in the docs. We could get around this by having these examples have a prefix (I guess `results`) and that prefix would be skipped in that test. Or remove the names from these chunks.

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -291,6 +293,22 @@ res.list[['lfc2']] <- list(
 )
 ```
 
+```{r assemble_variables, dependson=c('results', 'results2'), cache=TRUE}
```
daler, line 296: see comment above about naming. Also, if the docs-checking code is working properly, we'll get an error because this doesn't have a corresponding `assemble_variables` entry in the docs.

workflows/rnaseq/downstream/rnaseq.Rmd
```diff
@@ -291,6 +293,22 @@ res.list[['lfc2']] <- list(
...
+# NOTE: Add a new dependson for every contrast chunk to prevent this from
+# malfunctioning.
+# res.list is a named list. Each item should be a list with names c('res',
+# 'dds', 'label'). "res" is a DESeqResults object, "dds" is the corresponding
+# DESeq object the results were extracted from, and "label" is a nicer label to
+# use for headers and other text.
```
daler, line 302: this should go in those docs too

daler (commented):
Very cool idea!

workflows/rnaseq/downstream/rnaseq.Rmd
daler: @njohnso6 I think you can use `dependson=knitr::all_labels()[grepl('results', knitr::all_labels())]` to handle this automatically.

PR #350, opened 2021-10-05
Functional Enrichment Valid Key Exception Try-Catch

fridells51:
Currently, passing a nonempty list of non-protein-encoding genes that are absent from the data base to the functional enrichment portion of the rmd will result in an error that will interrupt the downstream analysis. 

Added a try-catch block to catch the exception and continue the analysis. When caught, the exception is noted in the rnaseq.html.


daler:
@fridells51 turns out the test fix was already implemented in v1.9rc, so I merged it in here.

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -29,10 +29,50 @@ id.convert <- TRUE
 # Convert gene IDs to those useable to clusterProfiler.
 # Docs: https://lcdb.github.io/lcdb-wf/rnaseq-rmd.html#bitrgenes
 # NOTE: Gene IDs are provided in what format?----------------------------------
-from.type <- "FLYBASE"
+from.type <- 'ENSEMBL'
+annotation_genus_species <- 'Mus musculus'
```
daler, line 33: the example data is from fly, so this should be changed back.

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -48,10 +88,14 @@ pvalueCutoff <- 1
 qvalueCutoff <- 1
 go.uni <- NULL
 path.uni <- NULL
-kegg.organism <- 'dme'
+kegg.organism <- 'mmu'
 RUN.REACTOME <- FALSE
-reactome.organism <- 'fly'
+reactome.organism <- 'mouse'
```
daler, line 93: same here, should go back to fly

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -29,10 +29,50 @@ id.convert <- TRUE
...
+from.type <- 'ENSEMBL'
+annotation_genus_species <- 'Mus musculus'
+annotation_key_override <- NA
+hub.cache <- '../../../include/AnnotationHubCache'
+orgdb <- get.orgdb(
+    annotation_genus_species,
+    cache=hub.cache,
+    annotation_key_override=annotation_key_override
+)
+# In place of AnnotationDbi::select, Use a wrapper function to run this command in order to catch the unmatching keys error
```
daler, line 41: I think this would fit better over in `lib/lcdbwf/R/helpers.R`. In fact, the `select.keyhander` could itself be created over there.

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -29,10 +29,50 @@ id.convert <- TRUE
...
+            }
+        )
+    }
+    wrapped_select
+}
+select.keyhandler <- keyhandler(AnnotationDbi::select, 'None of the keys')
 
 deg.list <- nested.lapply(sel.list, rownames)
-genes.df <- nested.lapply(deg.list, function (x) AnnotationDbi::select(orgdb, keys=x, columns=types, keytype=from.type))
+genes.df <- nested.lapply(deg.list, function (x) select.keyhandler(orgdb, keys=x, columns=types, keytype=from.type))
```
daler, line 73: if the function is moved to the lcdbwf package, then this would become:

```suggestion
genes.df <- nested.lapply(deg.list, function (x) lcdbwf::select.keyhandler(orgdb, keys=x, columns=types, keytype=from.type))
```

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -48,10 +88,14 @@ pvalueCutoff <- 1
 qvalueCutoff <- 1
 go.uni <- NULL
 path.uni <- NULL
-kegg.organism <- 'dme'
+kegg.organism <- 'mmu'
 RUN.REACTOME <- FALSE
-reactome.organism <- 'fly'
+reactome.organism <- 'mouse'
+min.gene.set.size <- 50
```
daler, line 94: I think this is a reasonable number and idea, but should be included in the PR (and eventually added to the changelog as well).

daler (changes requested):

daler:
Closing this since the downstream RNA-seq code has changed substantially in v1.9rc. @fridells51, when you get a chance could you submit a PR for this try/catch within the context of the new v1.9 code?

PR #354, opened 2021-12-24
Adding Samtools Idxstats to Snakefile RNAseq

fridells51:
Now, idxstats will be viewable in multiqc html after running the Snakefile. Added a new rule to the Snakefile and a new pattern to rnaseq_patterns.yaml.


daler:
@fridells51 FYI, I changed this PR to target `v1.9rc` instead of master, and merged recent test-fixing changes from v1.9rc into this branch.

daler:
@fridells51 would you please add this to the ChIP-seq workflow as well? While you're at it, I'd prefer the rule name `idxstats` rather than `sam_idxstats`.

daler:
Thanks!

PR #356, opened 2022-01-13
Rna rmd fixes

fridells51:
In the rnaseq.Rmd there is an issue where a command from BiocParallel is used before the BiocParallel library is loaded. This can be fixed by wrapping the method call, or by loading the libraries earlier in the file. I chose to do the latter.


daler:
@fridells51 can you remove the remaining libraries chunk?

daler:
Closing; these changes have since made it in to v1.9

daler:
Reopening - after a double check these have *not* made it in! Will merge into v1.10.

Issue #359, opened 2022-01-18
Refactor contrast sorting for output datatable in v1.9rc


kcampbell824:
v1.9rc current behavior sorts contrasts alphabetically by contrast name when summarizing DE results in the datatable presented in the rnaseq.html. Aim is to allow more flexibility in how contrast results are presented, for example in order of contrast creation. To do so, we will change the pattern for contrast names from `contr_` to `contr_NN_`, where NN is a numeric ordering of the contrasts. This name pattern will then sort the contrasts in order based on the NN index. Finally, we'll strip the `contr_NN` prefix from contrast names before adding to presented datatable.

kcampbell824:
Closed with merge of branch #361. Note that final version allows alphanumeric strings as sorting index.

PR #361, opened 2022-01-19
require contrast names include sorting index

kcampbell824:
Previously, v1.9rc required contrast names to follow the format `contr_[name]`, where [name] was used to alphabetically sort the output summary tables. To add flexibility, this branch requires the format `contr_NN_[name]`, where NN is a two-digit numeric index used to sort the contrasts. Users can modify NN to alter the order in which the contrast appears in output data tables. Documentation & example contrasts are update to reflect this change, as well. Addresses Issue #359

### DIFF
--- a/docs/rnaseq-rmd.rst
+++ b/docs/rnaseq-rmd.rst
@@ -268,8 +268,10 @@ its own chunk **according to the following rules**:
 - the chunk name must start with ``results_``
 - the chunk is cached
 - the chunk depends on the ``'dds_list'`` chunk
-- the variable name starts with ``contr_``, and the rest of the variable name
-  will be used as the name in the list
+- the variable name starts with ``contr_[index]_``, and the rest of the variable name
+  will be used as the name in the list. index is a string, containing 1 or more alphanumeric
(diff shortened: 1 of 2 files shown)


daler:
To support the ability to insert results in between existing sorted ones in a way that minimizes cache invalidation, I think it would be helpful to allow arbitrary strings in between the `_` to use when sorting. So e.g., if you have `contr_01_wt` and `contr_02_hom` but wanted to insert one in between, you could use `contr_01a_het` without needing to change the other names (which would invalidate their respective caches). I guess the only criteria would be that underscores cannot be allowed in that sorting label.

kcampbell824:
Most recent commits address @daler's suggestions. The index string can now contain alphanumeric characters. The only restrictions are that the index string must be at least length 1, to exclude contrast names like `contr__test`, where the index string is an empty string. We also require that the index string cannot contain "_". Docs have been updated to explain the new contrast naming conventions.

daler:
Looks great, thank you!

daler:
(and good call with specifying those additional constraints)

PR #362, opened 2022-01-20
Parallelize functional enrichment

mitraak:
- This PR adds the option to run the functional enrichment analysis in parallel using `bplapply`.
- Non-parallel operation uses `lapply` which still offers a slight improvement to the previous loop.
- Also adding KEGG database as an option for functional enrichment


daler:
@fridells51, @kcampbell824 when you identify a good solution for the cnetplot error,

```
Error in UseMethod("rescale") : 
  no applicable method for 'rescale' applied to an object of class "AsIs"
```
(i.e., when all GO terms have exactly one gene), please use this branch and push your fix here.

Note to future self: I temporarily modified the test config to only run the rnaseq test to save on resources while debugging this. The rest of the tests will need to be re-enabled before merging.

lib/lcdbwf/R/functional_enrichment.R
```diff
@@ -1,3 +1,106 @@
+#' Function to run enrichment on a res_list
```
daler, line 1: @mitraak, a lot of the code elsewhere has changed since this PR, making it a pretty messy diff. Can you pull out just this parallel part, make a branch off of `v1.10` to add it to, and open a new PR targeting `v1.10`?

Issue #365, opened 2022-05-17
ranseq.Rmd: Libraries loaded after function is referenced if parallel option is set in the config (v1.9)


fridells51:
If `parallel: TRUE` is set in the config, this code will be run in rnaseq.Rmd:

```
parallel <- config$parallel$parallel
if (config$parallel$parallel){
     register(MulticoreParam(config$parallel$cores))
}
```

The problem is that `register` comes from a library that is loaded downstream of this code block. Moving the library loading chunk up before this code runs will fix this

daler:
This is being addressed in #356

Issue #351, opened 2021-10-13
need index for markduplicates


daler:
We are aiming to have the markdups bam be the only one remaining; currently it does not have `.bai` created as part of the standard workflow.

PR #366, opened 2022-09-19
v1.10

daler:
- Refactored testing into `test/lcdb-wf-test` script, which makes it much easier to run local tests
- Raise exception if `include`ed reference config does not exist (#368)
- Require `layout` column when using SRA sampletable (#367)
- Ensure all libraries are loaded before configuring parallelization (#365)
- All `lcdbwf` R functions use the `:::` namespace lookup
- R functions and args follow style favoring `_` rather than `.` in var names
- Remove redundant libsizes table
- Ensure .bai exists for dups-marked ChIP-seq bams (#351)
- Refactor snakefiles to use `resources:` directive, and correspondingly stop using the (long-deprecated) `--clusterconfig` mechanism. Resources have been copied over from the clusterconfig. Use `LCDBWF_SNAKEMAKE_PROFILE` env var to point to a central profile directory
- Lots of updates for CI and making everything work with mamba
- Refactored deploy script to rely on distutils rather than rsync exclude (which appears broken in recent versions)
- Support specification (or not) of orgdb for getting annotation db (useful for genomes with no or limited annotations)
- New function `lcdbwf:::lfc_scatter` for comparing multiple DESeq2 runs
- Updates and fixes to `gene-patterns.Rmd` 
- remove clusterconfig.yaml files and cluster_specific.py
- update docs


workflows/rnaseq/Snakefile
```diff
@@ -138,7 +138,7 @@ if 'Run' in c.sampletable.columns and sum(c.sampletable['Run'].str.startswith('S
             sampletable=_st,
             # limit = 100000, # [TEST SETTINGS]
         resources:
-            mem_mb=1024,
+            mem_mb=autobump(1024),
```
daler, line 141: this should probably be changed to `disk_mb` for larger fastq files

include/WRAPPER_SLURM
```diff
@@ -7,6 +7,22 @@
 # make logdir
 if [[ ! -e logs ]]; then mkdir -p logs; fi
 
+# use SNAKEMAKE_PROFILE if one is set in the environment variables
+if [ -z "$LCDBWF_SNAKE_PROFILE" ]; then
```
daler, line 11: let's change to `LCDBWF_SNAKEMAKE_PROFILE` just to be consistent with the original name

workflows/rnaseq/Snakefile
esnaultc: done.

include/WRAPPER_SLURM
esnaultc: changed everywhere

Issue #377, opened 2023-07-06
Adding Epic2 as a new peakcaller in chipseq


Mira0507:
Epic2 has been introduced as a peakcaller for broad peaks. I'm adding a new rule to run epic2 in the `chipseq` workflow.

References:
- paper: https://pubmed.ncbi.nlm.nih.gov/30923821/
- github: https://github.com/biocore-ntnu/epic2

The working branch is `epic2` and will be merged into `v1.10.3rc` once it's completed.

Mira0507:
Closing.

PR #380, opened 2023-07-13
Epic2

Mira0507:
https://github.com/lcdb/lcdb-wf/issues/377


env.yml
```diff
@@ -1,7 +1,7 @@
-name: null
 channels:
   - conda-forge
   - bioconda
+  - defaults
```
daler, line 4: let's keep the defaults channel out -- it has historically caused some errors

```suggestion
```

include/reference_configs/Homo_sapiens.yaml
```diff
@@ -91,6 +91,7 @@ references:
         indexes:
           - 'salmon'
           - 'kallisto'
+      blacklist: 'https://github.com/Boyle-Lab/Blacklist/blob/master/lists/hg19-blacklist.v2.bed.gz'
```
daler, line 94: Can you add some documentation about this (what it does, how to get it for species of interest)?

workflows/chipseq/Snakefile
```diff
@@ -684,7 +726,8 @@ rule bed_to_bigbed:
 
 
         # bedToBigBed doesn't handle zero-size files
-        if os.stat(input.bed).st_size == 0:
+        # bigbed is not created from epic2-generated peaks
+        if os.stat(input.bed).st_size == 0 or "epic2" in input.bed:
```
daler, line 730: Not clear why we don't get a bigbed from epic2's bed. Does the bed file it creates not work with this rule? If that's the case we should fix it, possibly in a wrapper if needed (if the rule gets too messy/complicated).

workflows/chipseq/Snakefile
```diff
@@ -624,6 +624,48 @@ rule macs2:
...
+        runtime=autobump(hours=2)
+    log:
+        c.patterns['peaks']['epic2'] + '.log'
+    params:
+        block=lambda wc: chipseq.block_for_run(config, wc.epic2_run, 'epic2'),
+        genome=config['genome']
+    run:
+        # `-c` has to be skipped if no control is provided
+        if len(input.control) > 0:
+            cmd = '-c {input.control} '
```
daler, line 657: Since `cmd` is at most 2 args rather than the whole command, I think it would make more sense to change `cmd` to `arguments`

workflows/chipseq/Snakefile
```diff
@@ -624,6 +624,48 @@ rule macs2:
...
+        else:
+            cmd = ''
+        # Add `--guess-bampe` if input dataset is paired-end
+        if c.is_paired:
+            cmd += '--guess-bampe '
+        shell(
+            'epic2 ' + cmd +
```
daler, line 664:
```suggestion
            'epic2 '
            '{cmd} '
```

daler (changes requested):

include/reference_configs/Homo_sapiens.yaml
Mira0507: Thank you for pointing it out. I'm deleting it. My current update doesn't have blacklist filtering.

workflows/chipseq/Snakefile
Mira0507: I think the `bedToBigBed` cannot handle the format of epic2-created `bed` files.

* macs2

```bash
chr2L   5208    6064    gaf-wingdisc-pooled_peak_1      198     .       10.3022 22.4865 19.8284 464
chr2L   7465    7612    gaf-wingdisc-pooled_peak_2      23      .       3.56078 4.31716 2.38755 67
```

* epic2 

```bash
chr2L   5000    12399   1.59536939537213e-31    181.4181186791179       .       123     37      3.464230687093768e-301.814181186791179
chr2L   65000   83599   1.0311678411768225e-52  146.05718430892273      .       307     118     3.13475023717754e-51 1.4605718430892272
```

One way of resolving it is to convert epic2-bed to `bedToBigBed` compatible format using the `pybedtools`.

Mira0507:
@daler I mistakenly merged this before asking your review. Should I do the "revert"? Could you let me know what I am supposed to do in this case?

Mira0507:
Thank you!!

Issue #378, opened 2023-07-12
Deploy script error


njohnso6:
When run with python deploy.py --clone --dest project --flavor rnaseq

fatal: not a git repository (or any parent up to mount point /gpfs)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Traceback (most recent call last):
  File "/gpfs/gsfs8/users/CARD_AA/users/johnsonnicl/test/deploy.py", line 392, in <module>
    include = write_include_file(source)
  File "/gpfs/gsfs8/users/CARD_AA/users/johnsonnicl/test/deploy.py", line 144, in write_include_file
    sp.check_output(
  File "/usr/local/Anaconda/envs/py3.10/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/local/Anaconda/envs/py3.10/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'ls-tree', '-r', 'HEAD', '--name-only']' returned non-zero exit status 128.

Edit: turns out it is important to run the deploy script from within a repo that is already a git repo. If you do it works fine. But that isn't clear in the documentation.

PR #384, opened 2023-07-27
378 deploy script error

aliciaaevans:
- Fix https://github.com/lcdb/lcdb-wf/issues/378 by passing in the source dir
- `flavor `wasn't properly being passed in and compared, so all workflows were being deployed
- The `--staging` default value was causing it to be considered always set. Instead of using a default on the arg, only use the default if `--clone` is set.
- improve documentation and comments in `test/lcdb-wf-test` and `deploy.py`


deploy.py
```diff
@@ -107,11 +107,11 @@ def write_include_file(source, flavor=None):
...
 
     patterns = []
-    if flavor is None or 'rnaseq':
+    if flavor is None or flavor == 'rnaseq':
         patterns.extend(PATTERN_DICT['rnaseq'])
-    if flavor is None or 'chipseq':
+    if flavor is None or flavor == 'chipseq':
         patterns.extend(PATTERN_DICT['chipseq'])
-    if flavor is None or 'full':
+    if flavor is None or flavor == 'full':
```
aliciaaevans, line 114: @daler Should full include chipseq and rnaseq?

daler: ah, yep, just pushed an attempted fix

daler:
@aliciaaevans OK to merge this?

aliciaaevans:
> @aliciaaevans OK to merge this?

@daler yes, everything looks good to me.

PR #394, opened 2023-11-28
Added Error header in rnaseq-rmd.rst and docs for 'Found duplicate na…

### LINKED ISSUE #393: Add documentation for "Found duplicate names after removing pattern ^contr_[^_]+_" error
**Problem:**

In _rnaseq.Rmd_, there are many ways to attempt to add a previously existing contrast name to res_list. By accidentally attempting to do so, users encounter a single error that may be misleading in some cases. Here is an example of such a case:

1. User sets up contrasts and runs all results_## chunks with config=TRUE, as the default examples show.
2. Invalidates the cache of at least one results_## chunk.
3. Re-renders the .Rmd file.
4. The results_## chunk with invalidated cache runs again, and there is an attempt to add an object with an already been used name to res_list. This triggers a false duplicate contrast name error, as the mechanism checking for duplicates does not recognize if the same contrast is being rerun or if a new object with an already been used name is being added to res_list.
5. The user encounters the error message: "Found duplicates after removing the pattern ^contr_[^_]+_", which is misleading in this case where there are no visible duplicate contrast names.

**Solution:**

Update the documentation to clarify that rerunning a results_## chunk will trigger the Found duplicates error. 

Improve the error message to include: _"Don't have duplicate contrast names? Read here: https://lcdb.github.io/lcdb-wf/rnaseq-rmd.html#downstream-detailed#duplicates-found-after-removing"_

menoldmt:
…mes' error in rnaseq.Rmd closes #393

### DIFF
--- a/docs/rnaseq-rmd.rst
+++ b/docs/rnaseq-rmd.rst
@@ -544,6 +544,41 @@ enrichment RMarkdown document.
 
 The output of sessionInfo records the versions of packages used in the analysis.
 
+Errors
+------
+
+In this section, we address some common errors encountered during the RNA-Seq downstream analysis
+and provide guidance on how to resolve them.
+
+Error: "Found duplicate names after removing pattern ^contr_[^_]+_"
+NOTE: Error in:: purrr::map() can also be caused by this:
+   Cause:
+      - If there are no duplicate contrast names in the:: results_## chunks:
+        a more subtle cause for this error occurs when the pattern is altered
+        on a previously cached chunk. Even if the change is undone, the
+        environment is cleared, and the file is rerun, the error will persist.
+        An example of this would be if you completed the:: results_01 chunk
+        followed by the:: results_02 chunk (with both chunks cached), rendered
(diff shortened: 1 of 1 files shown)


docs/rnaseq-rmd.rst
```diff
@@ -544,6 +544,41 @@ enrichment RMarkdown document.
...
+        unable to differentiate between the old and new contr_xx patterns. You
+        may be wondering, "how is the old pattern persisting in the
+        R environment after it has been changed and the file has been rerun
+        (even after quitting R/clearing the environment and workspace)." That
+        is because the:: results_02 chunk cache contains the former::
+        results_01 chunk's objects even after they have been changed. 
+
+   Solution:
+      - All cache and R environment objects must be cleared upstream of the::
+        assemble_variables chunk. To do so: Quit R (without saving the
```
daler, line 578:
```suggestion
      - All cache and R environment objects must be cleared upstream of the
        ``assemble_variables`` chunk. To do so: Quit R (without saving the
```

PR #399, opened 2024-05-08
Generate RDS object ready for carnation

mitraak:
This PR contains a rewrite of the `compose_results` function, which is now configured to
accept an RDS file path or multiple object lists and output an RDS object ready to plug into
carnation.


lib/lcdbwf/R/helpers.R
```diff
@@ -473,9 +479,14 @@ nested.lapply <- function(x, subfunc, ...){
 #'     ...
 #'   )
 #'
-#' `enrich_list` is optional. Note that `enrich_list`, if provided, must use
-#' results names available in `res_list`. In this example, the names are
-#' "ko.vs.wt" and "het.vs.wt". It has the following format:
+#' `enrich_list` is optional. Note that `enrich_list`, if provided, these are usually
+#' results names available in `res_list`. Alternatively, can have a 'res' key at the
+#' second-level containing a result name available in `res_list`.
```
daler, line 484:
```suggestion
#' `enrich_list` is optional. If provided, has names corresponding to
#' results names in `res_list`. Alternatively, can have a 'res' key at the
#' second-level containing a result name available in `res_list`.
```

Is there a historical reason we need to support both ways of connecting to a results object?

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
+    }
+
+    res_list <- tmp$res_list
+    dds_list <- tmp$dds_list
+
+    if('rld_list' %in% names(tmp)) rld_list <- tmp$rld_list
+    if('enrich_list' %in% names(tmp)) enrich_list <- tmp$enrich_list
+    if('degpatterns_list' %in% names(tmp)) degpatterns_list <- tmp$degpatterns_list
+    if('all_dds' %in% names(tmp)) all_dds <- tmp$all_dds
+    if('all_rld' %in% names(tmp)) all_rld <- tmp$all_rld
```
daler, line 549: Would be useful to add to documentation that whatever's in the RDS file overrides what's provided to this function. Actually, now that I think about it, it might be good to flip it around (whatever is provided to the function wins). That way you have the option of overriding or not. Currently you can't opt to replace what's in the RDS.

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
   dds_dds_names <- names(dds_list)
   res_not_dds <- setdiff(res_dds_names, dds_dds_names)
   dds_not_res <- setdiff(dds_dds_names, res_dds_names)
   if (length(res_not_dds) > 0){
-    stop(paste("The following dds names are in res_list but are not found in dds_list:", res_not_dds, '\n'))
+    stop(paste("\t- The following dds names are in res_list but are not found in dds_list:",
+               paste(res_not_dds, collapse=', '), '\n'))
   }
+
+  # NOTE: drop unused dds_list & rld_list objects
```
daler, line 564:
```suggestion
  # Drop unused dds_list & rld_list objects
```

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
+                          rld_list=rld_list)
+
+  # if all_dds not specified, but dds_list has length 1,
+  # then use dds_list[[ 1 ]] as all_dds
+  if(is.null(all_dds) & length(obj$dds) == 1){
+    message('\t- all_dds was not specified, but dds_list has only 1 object. Using that instead')
+    all_dds <- obj$dds[[ 1 ]]
+
+    # if specifying all_dds, compute all_rld even if specified
+    all_rld <- varianceStabilizingTransformation(all_dds, blind=TRUE)
```
daler, line 594: Not obvious to me why it needs to be remade, I would think if you had a single dds you'd have a single rld? Either way, should at least give a message here because I would not expect a function to silently overwrite an arg I had provided.

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
+    gidx <- which(tolower(colnames(res)) %in% 'gene')
+
+    gene2symbol <- c(gene2symbol, unname(res[, sidx]))
+    gene2symbol_names <- c(gene2symbol_names, res[, gidx])
+  }
+
+  # remove duplicates
+  idx <- !duplicated(gene2symbol)
+  gene2symbol <- gene2symbol[!idx]
+  names(gene2symbol) <- gene2symbol_names[!idx]
```
daler, line 620: Not sure if it's cleaner to use a list and overwrite each time, like:

```r
gene2symbol[res[, sidx]] <- res[, gidx]
```

(saves on the deduping step)

Also, are there cases where we'll be missing genes in some results objects, such that this needs to be done for every results object (rather than picking the first one)? Maybe if there's an edgeR-type filtering step to remove low-count genes or something. OK, I convinced myself that yes, checking all (and then using the lookup in the lines below) is the most robust approach.

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
+        for(z in names(enrich_list[[x]][[y]])){
+          elem_names <- c(elem_names, paste(x, y, z, sep=sep))
+        }
+      }
+    }
+    names(elem_names) <- elem_names
+
+    message(paste('\t\t- Running conversion using', workers, 'worker(s)'))
+    # run conversion
+    # TODO: add check for cores if on biowulf
```
daler, line 688:
```suggestion
```

lib/lcdbwf/R/helpers.R
```diff
@@ -493,59 +504,413 @@ nested.lapply <- function(x, subfunc, ...){
...
+#' - Builds a dds_mapping object that maps res_list objects to dds_list objects
+#' - Flattens res_list object which is replaced by two slots corresponding to
+#'   'res' & 'label' elements.
+#'
+#' @param res_list List of DESeqResults objects
+#' @param dds_list List of dds objects
+#' @param rld_list List of normalized dds objects
+#' @param reserved_cols Column names reserved for internal use. colData
+#'        of dds_list or rld_list objects cannot contain these columns
+#'
```
daler, line 814:
```suggestion
#' @return A list with the names 'res', 'dds', 'rld', 'labels', 'dds_mapping'
```

env-r.yml
```diff
@@ -2,403 +2,432 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
```
daler, line 4: Can you remake the `env-r.yml` using `--no-builds`? See [here](https://bioconda.github.io/faqs.html#why-shouldn-t-i-include-build-numbers-in-my-environment-yamls) for rationale.

daler (changes requested):
This is great! I especially appreciate all the error checking and handling of corner cases.

At some point we should add to the tests to make sure all of this is functioning properly (and to make sure it continues to do so in the future as things change).

lib/lcdbwf/R/helpers.R
mitraak: I agree. I think it would be intuitive to have specified arguments win over what's in the RDS file. Will add that

lib/lcdbwf/R/helpers.R
mitraak: Agreed, there should at least be a message. Maybe `all_rld` shouldn't even be an argument, i.e. the function
will always compute this. I basically have this here to ensure a 1:1 mapping between `all_dds` and `all_rld`, which
might not be true if i allow independent objects to be specified. Basically, this protects from user error.

env-r.yml
mitraak: Thanks, I noticed no builds in lcdb-wf and was wondering how to get that. Will add

lib/lcdbwf/R/helpers.R
daler: I don't have a preference for what should be default (or what would objectively be the "best"). Though I feel like a separate enrich_mapping would be the most flexible.

mitraak: Yes, agreed. A separate enrich_mapping would be nice, but will need some tweaking in carnation.
Will add that to a future update.

workflows/rnaseq/downstream/gene-patterns.Rmd
```diff
@@ -22,6 +22,7 @@ library(tidyr)
 library(clusterProfiler)
 library(DESeq2)
 library(DEGreport)
+library(GeneTonic)
```
daler, line 25: is this needed in gene-patterns.Rmd?

daler (changes requested):
So if I understand things correctly, functional-enrichment.Rmd and gene-patterns.Rmd can run in any order, is that right? We should probably get some tests going to verify that all works correctly

Issue #400, opened 2024-05-10
Adding universe to functional enrichment analysis


Mira0507:
The `universe` argument will be used to limit background genes only to detected genes in input dataset, when calling the `enricher` function in `functional-enrichment.Rmd`. 

See https://github.com/YuLab-SMU/clusterProfiler/blob/devel/R/enricher.R#L8C21-L8C136 for the original usage.

Mira0507:
Closing.

PR #401, opened 2024-05-17
Limit background genes in overrepresentation analysis in functional-enrichment.Rmd

Mira0507:
This pull request is opened to cover the issue described in https://github.com/lcdb/lcdb-wf/issues/400.


lib/lcdbwf/R/functional_enrichment.R
```diff
@@ -101,11 +103,14 @@ collapse_names <- function(res_list, config, sep='*'){
 #'   lfc_thresh from the config.
 #' @param kind One of "OR" for overrepresentation or "GSEA" for gene set
 #'   enrichment analysis.
+#' @param universe background genes. If missing, all genes listed in the database 
+#'   (e.g. TERM2GENE table) will be used as background. See more info in
+#'   https://github.com/YuLab-SMU/clusterProfiler/blob/devel/R/enricher.R#L8
```
daler, line 108: This line may change over time as clusterProfiler develops. That said, this line 8 says the same thing this help string says, so I don't think the link is needed.

workflows/rnaseq/downstream/config.yaml
```diff
@@ -236,3 +236,7 @@ functional_enrichment:
   # fine getting plots even with no statistically significant terms.
   pvalueCutoff: 1
   qvalueCutoff: 1
+
+  # This sets the `universe` argument of the `enricher` function to only detected genes
+  # See https://github.com/YuLab-SMU/clusterProfiler/blob/devel/R/enricher.R#L8C21-L8C136
```
daler, line 241: Similar comment to above, I don't think the link is helpful.

lib/lcdbwf/R/functional_enrichment.R
```diff
@@ -121,13 +126,19 @@ enrich_test <- function(res, TERM2GENE, TERM2NAME, config, direction, kind='OR',
       direction=direction,
       return_type="rownames"
     )
-
+    # Update the `universe` variable to a vector of background or all genes
+    if (universe) {
+      universe <- rownames(res$res) # uses only detected genes in the current dataset
```
daler, line 131: For this line, if I understand correctly, "detected" is being defined as "everything in the results object" which ultimately corresponds to, "every annotated gene seen by featureCounts".

Or is the idea that you would manually subset `res$res` before sending it here to only include, say, `res$res[rowSums(res$res)>0,]`?

Maybe we need a way of specifying a list of genes to be passed in here. Or maybe do the `rowSums > 0` within this function? Or perhaps filter out anything with [`isna(padj)`](https://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#pvaluesNA) to get rid of genes that were not even considered by multiple testing correction?

daler (changes requested):
Thanks! I'd like to kick around some ideas of how best to either filter to actual detected genes, or to provide a straightforward mechanism for selecting the universe (perhaps a `detected()` convenience function?)

lib/lcdbwf/R/functional_enrichment.R
Mira0507: ok, I'm deleting the url.

workflows/rnaseq/downstream/config.yaml
Mira0507: deleted!

lib/lcdbwf/R/functional_enrichment.R
Mira0507: My design was the same as what's been described here:

>  "detected" is being defined as "everything in the results object" which ultimately corresponds to, "every annotated gene seen by featureCounts".

Assume there are 8 samples: WT rep1-4, KO rep1-4. Then, there are 3 gene groups here:

a. genes with raw counts > 0 in all samples
b. genes with raw counts > 0 in some samples 
c. genes with raw counts > 0 in no samples

_Raw counts_ mean the number of read counts annotated by featureCounts and my `if` statement was designed to choose genes belonging to 'a' and 'b'. Checking [DESeq2 documentation](https://bioconductor.org/packages/3.7/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#pvaluesNA), I assumed that the `baseMean` and `log2FoldChange` columns would return `NA` if a gene is not counted in all samples ('c'). At least within my projects, no `NA` was found in the `baseMean` and `log2FoldChange` columns. 

Regarding this issue, I can bring up two scenarios:

- scenario 1: genes with `NA`s in the `baseMean` and `log2FoldChange` columns were already removed before being saved in the `combined.Rds` file in the `rnaseq.Rmd` script
- scenario 2: coincidently my dataset didn't have `NA`s in those two columns

I thought this case was the scenario 1 but I couldn't find the filtering step when looked into `rnaseq.Rmd`, `helpers.R`, `contrasts.R`, and `functional-enrichment.R`. It's still unclear. Adding a filtering step sounds like a great idea if filtering was not performed in the previous script (e.g. `rnaseq.Rmd`).

Comparing total detected genes vs genes considered by multiple hypothesis testing, I would like to go with total detected genes, because they will be used as _background_ of overrepresentation analysis, conceptually. Here, I considered biological background rather than statistical background. I'd like to hear your thoughts! :)

Mira0507:
When it comes to this conflict, I see a difference in the `dependson` option. FYI, I created my `universe` branch from `master` and targeted the `v1.11` branch for my pull request. Could you clarify what version of `dependson` should be kept?

lib/lcdbwf/R/functional_enrichment.R
daler: I think my broader question is, should the `universe` argument be a boolean or a list of genes?

Currently it's bool, so it's up to the user to construct the results object appropriately, and that should be more clearly indicated in the help for the function as well as in the config file.

If `universe` should instead be a list of genes, then a helper function would be nice, and that could allow the user to choose what "background" means: any genes with nonzero counts in any sample, or any genes with a valid `padj` (which would include genes with zero counts as well as those with too-low to consider).

I'm leaning towards the list-of-genes approach because it's more explicit. However it would need some more thinking about how to handle it in the config file. Maybe the `limit_background_genes: TRUE` should instead allow different methods of choosing background? And if FALSE then the default universe (all annotated genes) is used.

daler:
I fixed the conflict -- I had recently merged in #399, soyou may have had an out-of-date version of v1.11 branch at the time you made your new branch.

lib/lcdbwf/R/functional_enrichment.R
Mira0507: I got your point. It sounds reasonable to me, too! 

How do these 4 categories look in the `config.yaml` file? 

```yaml
  # This sets the `universe` argument of the `enricher` function to choose an option how to limit
  # background genes. Choose one from the following options:
  # - positive_counts_all: genes with positive raw counts in all samples
  # - positive_counts_some: genes with positive raw counts in one or more samples
  # - valid_padj: genes analyzed in multiple hypothesis testing. In DESeq2, genes with very
  #   low counts have high dispersion. These genes are not considered to be significantly different
  #   between groups, and labeled NA in padj.
  # - all_in_DB: disabling to limit background genes. This option will use all available genes in each DB.
  limit_background_genes: "positive_counts_all"
```

I have terrible naming sense. Feel free to suggest better names. 
If they look ok, I'll go related functions based on them.

daler: How about the following? I tried to be very clear about what's being excluded. Also, since the background should be "anything that had a chance of being detected", I think the NA/padj business is too complex, and so removed that. Then, I couldn't think of a situation where I would want to exclude genes with a zero in *any* sample, which seems pretty restrictive.

So that brings us down to two options, but with the flexibility to include more options later if it makes sense.

```yaml
  # This sets the `universe` argument of the `enricher` function to choose an option how to limit
  # background genes. Choose one from the following options:
  #   - exclude_zero_in_all: excludes genes with zero raw count in all samples
  #   - no_filter: disable any filtering. This option will use all available genes in each annotation database. You can always use this option and then manually adjust the results objects going into the `enricher` function if you need more customization.
  limit_background_genes: "exclude_zero_in_all"
```

But now that's down to only two options, I'm not sure if it makes sense to have this in the config at all. Furthermore, since the `enrich_test` function already passes its `...` args to `enricher` and `GSEA`, we can already do this in the master branch:

```r
enrich_test(res, TERM2GENE, TERM2NAME, config, direction='up', universe=names_of_genes_to_use)>0])
```

So maybe the solution here is: 1) by default, restrict to nonzero; 2) make it obvious that's what's happening; and 3) possibly provide a helper function to do this.

E.g., 

```r
nonzero_genes <- function(res){
  # something like this
  return names(res)[res$baseMean == 0]
}
enrich_test(res, TERM2GENE, TERM2NAME, config, direction='up', universe_func=nonzero_genes)
```

Mira0507: `exclude_zero_in_all` and `no_filter` sound great. 

When it comes to the `enriche_test` function, I think we can keep it as-is as you described. The `enrich_test` is called in the `run_enricher` function. Therefore, one sensible way would be to narrow the `universe` using a wrapper function like your `nonzero_genes` and assign them as input genes before `enrich_test` in the `run_enricher()` function.

workflows/rnaseq/downstream/config.yaml
```diff
@@ -237,6 +237,11 @@ functional_enrichment:
...
-  # This sets the `universe` argument of the `enricher` function to only detected genes
-  # See https://github.com/YuLab-SMU/clusterProfiler/blob/devel/R/enricher.R#L8C21-L8C136
-  limit_background_genes: TRUE
+  # This sets the `universe` argument of the `enricher` function to choose an option how to limit
+  # background genes. Choose one from the following options:
+  # - exclude_zero_in_all: excludes genes with zero raw count in all samples
+  # - no_filter: disable any filtering. This option will use all available genes in 
+  #     each annotation database. You can always use this option and then manually 
+  #     adjust the results objects going into the `enricher` function if you need 
+  #     more customization.
```
daler, line 246: Given the above implementation where `run_enricher` accepts a list of vectors, this should be updated

lib/lcdbwf/R/functional_enrichment.R
```diff
@@ -44,7 +44,7 @@ run_enricher <- function(res_list, ontology_list, config,
               TERM2GENE=ontology_list[['term2gene']][[ont]],
               TERM2NAME=ontology_list[['term2name']][[ont]],
               config=config,
-              universe=universe
+              universe=universe_list[[name]]
```
daler, line 47: Should add a check earlier in this function to ensure that a universe entry exists in the list for all names in `res_list`, in cases where the user provides their own. If a name doesn't exist, I guess give an error? Or is there a sensible fallback behavior?

workflows/rnaseq/downstream/functional-enrichment.Rmd
```diff
@@ -71,16 +71,33 @@ if(!'degpatterns_list' %in% names(obj)){
...
+# Print the current universe option
+if (univ_option == "exclude_zero_in_all") {
+    # Retrieve all genes detected in one or more samples per contrast
+    universe_list <- lapply(names(dds_list), function(name) nonzero_genes(dds_list[[name]]))
+    names(universe_list) <- names(dds_list)
+} else if (univ_option == "no_filter") {
+    # Set universe to NULL
+    universe_list <- NULL
+} else {
+    stop("Correct your background option in the config.yaml file.")
```
daler, line 89: Might as well add the options to the error message

daler (changes requested):

lib/lcdbwf/R/functional_enrichment.R
Mira0507: The `universe_list[[name]]` returns `NULL` when `universe_list` was set to `NULL` in the previous chunk (when `config$functional_enrichment$limit_background_genes` being `no_filter`).

```r
> universe_list
NULL
> universe_list[['WT_Dox_vs_WT_Vehicle']]
NULL
```

Then, the `enrich_test` function calls `clusterProfiler::enricher` function where the `universe` argument is set to `NULL`, which will use all genes in each database. 

I will add a step raising an error when custom `universe_list` has different names from `res_list` if it's set by user manually.

daler:
Looks good, and the [built artifact from tests](https://output.circle-artifacts.com/output/job/b5b59e1f-2761-4aaa-813a-213935c4294c/artifacts/0/functional-enrichment.html) ran fine. Eventually we should write some proper tests for this, but I think this is good to go. Any other changes you want to make before merging?

PR #411, opened 2024-10-09
Fix unused r1_only parameter

b-full:
The render_r1_r2() function in the rnaseq and chipseq Snakefiles had an unused r1_only=False parameter which was removed


b-full:
Closed in favor of pull request #413

PR #412, opened 2024-10-09
Move strandednes in params

b-full:
Move the assignment of strand_arg into params for tools that need strandedness so that --rerun-trigger will detect changes to strandedness configuration and re-run those rules


daler:
Closing in favor of #415

PR #414, opened 2024-10-15
Make strand_arg a param

b-full:
Move `strand_arg` assignment from the `run` block to the `params` block so that `--rerun-trigger` will detect changes to strandedness configuration and re-run those rules


daler:
Closing in favor or #415

Issue #408, opened 2024-10-08
SRA runs should save data elsewhere and then symlink


daler:
Currently, when running an SRA metadata sample table, the fastq files are saved directly in the `data` directory. However, this means that you can't delete the `data` directory to trigger a re-run of a workflow like you can in a non-SRA run, where it's just the symlinks living in the data directory -- because the downloaded SRA fastqs would be deleted as well.

Instead, fastq-dump should put the files in a separate directory (will need to think carefully about where this should be), and then symlink to `data`.

PR #418, opened 2024-10-25
Change SRA fastq directory

b-full:
Change the directory where SRA fastq files are downloaded and add the 'orig_filename' column to the config object for each sample so that the rest of the workflow works correctly


b-full:
Addresses #408

Issue #407, opened 2024-10-08
Layout rather than LibraryLayout


daler:
Recently it seems that the SRA sampletables downloaded from SRA use `Layout` rather than `LibraryLayout` as a column name. This breaks the SRA detection. Confirm this is the case, and make the SRA detection more robust, for example by looking for either of those column names.

menoldmt:
Confirmed this is an issue. I was only able to run `./run_complex.test.sh -j16` after removing the LibraryLayout column in  `../../test/test_configs/complex-dataset-rnaseq-sampletable.tsv` and replacing it with `layout` with `SE` in all rows.

menoldmt:
My first error message after running `./run_complex_test.sh -j16` without editing the sampletable:

```
ValueError in file /gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/Snakefile, line 30:
Sampletable appears to be SRA, but no 'Layout' column found. This is required to specify single- or paired-end libraries.
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/Snakefile", line 30, in <module>
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/../../lib/patterns_targets.py", line 106, in __init__
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/../../lib/patterns_targets.py", line 78, in __init__
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/../../lib/helpers.py", line 20, in detect_layout
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/../../lib/helpers.py", line 20, in <listcomp>
  File "/gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/../../lib/common.py", line 732, in is_paired_end
```

my second error message after running `./run_complex_test.sh -j16` after adding `layout` (even though the error message has `Layout`):

```
ValueError in file /gpfs/gsfs10/users/NICHD-core0/test/menoldmt/expression-mod/workflows/rnaseq/Snakefile, line 30:
Expecting column 'layout' or 'LibraryLayout', not both
[...]

PR #419, opened 2024-11-08
Account for different 'layout's in sampletable

b-full:
-Implemented changes to account for 'LibraryLayout', 'Layout' and 'layout' formats in sampletable.
-Refactored the test suite and added pytest tests to check for correctness
-Added a helper function that recursively searches the directory structure and returns the path to the top level directory in an lcdb-wf repo


b-full:
Addresses #407

PR #425, opened 2024-12-13
Use tempdir, if configured, for STAR

njohnso6:
Have STAR use disk memory for its temporary files.


daler:
@njohnso6 tests are having an issue with `$TMPDIR`. CircleCI lets you log in via SSH (if owner of project); I did that and confirmed that `$TMPDIR` is not set by default. So the arg is now only added if that env var exists.

PR #431, opened 2025-02-27
V1.12rc

njohnso6:
Also included a change to memory usage for star because it has been giving me problems. This is with a new dataset of 6 files but the average size is 7.5GB per fastq. 
Also, will note that there is still a problem with the thread argument not working. Though thread numbers are indicated in the Snakefile they do not seem to affect the actual number of threads used.


daler:
Closing this since I'd like to keep cluster implementation details (like biowulf) out of this code.

